PR #1, opened 2014-08-19
small adjustments on path, db_port and servicegroup object

MelanieGault:
I made little adjustements 
- to have the default path in all objects reflecting the directories created in server.pp
- in the doc there is a db_port parameter for the server : I add it in server.pp definition and params.pp
- I need servicegroup object : I add it

I edited with gepetto with lint autocorrection activated so it has reformat every file.

Tested on centos


dnsmichi:
Could you please create a referencing issue at https://dev.icinga.org/projects/icinga-tools?jump=welcome so that we may coordinate and close the pull requests. Thanks.

MelanieGault:
I have no permission to sign in.

dnsmichi:
You'll need your Icinga account for that. If you don't have one, please register one (https://www.icinga.org/icinga/faq/). You can use that one for the wiki and future login forms on *.icinga.org too. 

Sorry for the inconvenience, but redmine is our external issue tracker for 5+ years now, and that won't change in the future.

MelanieGault:
redmine #6967 
I will continue and add some more objects I need tomorrow. I will do redmine tickets.

dnsmichi:
Ok, very nice, thank you!

MelanieGault:
I close this one and I will do as asked in https://dev.icinga.org/issues/6967 : 1 pull request for each correction.

PR #3, opened 2014-08-22
Object servicegroup

MelanieGault:
Add object service group.
I checked hostgroup/servicegroup typo everything should be fine.

relative to https://dev.icinga.org/issues/6967


nickchappell:
These were fixed/added in commits 52da440, 73190ad and b3ad2d0.

PR #4, opened 2014-08-22
object_user.conf.erb missed = on period parameter

MelanieGault:
object_user.conf.erb missed = on period parameter
add = to make it ok.

linked to : https://dev.icinga.org/issues/6983


nickchappell:
@dnsmichi, I've merged this. Could you close this PR?

PR #5, opened 2014-09-10
Various fixes

t-8ch:
Mostly to make it easier to support Debian in the future (We intend to use this module there)

### DIFF
--- a/manifests/nrpe/command.pp
+++ b/manifests/nrpe/command.pp
@@ -4,14 +4,14 @@
 #
 # Parameters:
 # * $command_name = What NRPE will know the command as; this defaults to the title of the resource
-# * $nrpe_plugin_liddir = The directory where the NRPE plugins themselves live
+# * $nrpe_plugin_libdir = The directory where the NRPE plugins themselves live
 # * $nrpe_plugin_name = The name of the plugin the command will run
 # * $nrpe_plugin_args = The arguments to pass to the plugin. This may be optional, 
 #                       depending on the plugin and whether it expects any arguments or parameters
 
 define icinga2::nrpe::command (
   $command_name       = $name,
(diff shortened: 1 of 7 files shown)


nickchappell:
Hi @t-8ch,

Sorry for the inconsistencies in Debian mentions and the typos. I was going to clean those up but have been really busy lately with work.

I should be able to remove the Debian mentions and any implications of Debian or Red Hat (the actual OS, not the OS family) support in the module tonight. I do want to support Debian and some other OSes (other EL derivatives, SuSE), but want to clean the module up first so me and any other contributors are starting from a cleaner base.

I know it's tedious, but would you be able to create separate PRs for the full qualification of fact variable names and the "liddir" typo? Those commits look good and I should be able to merge them as-is.

On the issues related to de-duplication and using `$::osfamily` instead of `$::operatingsystem`, I'd rather stick with using `$::operatingsystem` as different distros within the same family can pick different things in the packages or might require separate settings that I configure elsewhere in the module. It's arguably not too much of a problem now with just CentOS 6 and Ubuntu 12.04/14.04 being supported, but I'd like to keep the current structure in place to make future OS and OS version additions easier in the future. It is a fair amount of duplicated variables, but I don't think it's too bad if it's kept in params.pp as much as possible.

I can clean up things like the inconsistent use of the regexes when I look at the `$::operatingsystem` fact value, though.

The Icinga project uses its own issue tracker instead of Github's. Would you be able to make an account on https://dev.icinga.org/, open separate issues for each PR in the Icinga Tools project and assign them to me? We can do further discussion there.

t-8ch:
@nickchappell 
Thanks for the feedback, I'll do as you suggested.
`nagios-plugins-common` will be pulled in as a dependency and doesn't have to be specified.

PR #8, opened 2014-09-16
Added Debian 7 support

alvagante:


nickchappell:
Just for reference, I opened an issue on dev.icinga.org for this:

https://dev.icinga.org/issues/7264

dnsmichi:
I would appreciate it if everyone putting a pull request here would create a referencing issue on their own :) So we can organize any future milestones and coordinate event better. TIA. https://dev.icinga.org/projects/icinga-tools (registering an account is easy: http://www.icinga.org/register )

manifests/params.pp
```diff
@@ -102,6 +108,22 @@
       }
     }
 
+    #Debian systems:
+    'Debian': {
+      case $::operatingsystemmajrelease {
+        #Only tested on Debian7
+        '7': {
+          $icinga2_server_package = 'icinga2'
+          $icinga2_server_plugin_packages = ["nagios-plugins", "nagios-plugins-basic", "nagios-plugins-standard", "nagios-snmp-plugins", "nagios-plugins-extra", "nagios-nrpe-plugin"]
```
nickchappell: `nagios-plugins-extra` doesn't appear to be available on Debian 7. It does have `nagios-plugins-contrib` available, however.

manifests/params.pp
```diff
@@ -367,6 +357,19 @@
       }
     }
 
+    #Debian systems:
+    'Debian': {
+      case $::operatingsystemmajrelease {
+        '7': {
+          $icinga2_client_packages = ["nagios-nrpe-server", "nagios-plugins", "nagios-plugins-basic", "nagios-plugins-standard", "nagios-snmp-plugins", "nagios-plugins-extra", "nagios-nrpe-plugin"]
```
nickchappell: Same thing as above with the `nagios-plugins-extra` package not being available.

manifests/params.pp
alvagante: Right, sorry I didn't test the nrpe installation.
New PR arriving.

metadata.json
```diff
@@ -16,6 +16,12 @@
       ]
     },
     {
+      "operatingsystem": "Debian",
+      "operatingsystemrelease": [
```
cropalato: Maybe you can use operatingsystemmajrelease.

alvagante:
Ugh, fixed the metadata, but in the PR were added also a pair of commits for the setup of the classicui (they should be fine, but I'm not sure of the settings for $icinga_classicui_apache_conf_link and $icinga_classicui_apache_conf_path on Centos/Ubuntu).
Also it expects that you use apache which is somehow opiionated (but follows the same package's approach, which provides confiraton files for apache only)

nickchappell:
Hi @example42,

Would you be able to rebase/resubmit the PR without the commits for the classic UI setup? I was going to handle the installation/configuration of the UIs as a separate module or modules at some point in the near future.

alvagante:
Yet another commit added, since the PR points to my master... 
This is a generic icinga2::conf define that allows quick creation of custom config files in conf.d with custom templates (which can us variables for a custom options_hash) or static files

nickchappell:
Hi @example42,

I'd like to add the *.conf file define, but would you be able to submit that as a separate PR, and include a section in the README about it and how to use it?

I created an issue on the bug tracker: https://dev.icinga.org/issues/7273

alvagante:
Made  a new PR #9

nickchappell:
Thanks!

PR #17, opened 2014-09-22
Use debmon repository

albustax:
Hi,

This feature allow the use of [Debmon repository](http://debmon.org/packages) for Debian 7. You can use it like that:

<pre>
#Install Icinga 2:
class { 'icinga2::server':
  server_db_type => 'pgsql',
  # default to false
  use_debmon => true,
  db_host => 'localhost'
  db_port => '5432'
  db_name => 'icinga2_data'
  db_user => 'icinga2'
  db_password => 'password',
}
</pre>

[Debmon repository](http://debmon.org/packages) is up to date compared to the backports and regularly updated.

Have a good day,

Olivier


cropalato:
Even if debmon is up to date compared to the backports, this fix will fail if the icinga host do not have access to the internet. I am not sure if it should be inside the module, but maybe you can have  an option to define another repo (hiera level or something like this)

nickchappell:
Hi @albustax,

Would you be able to restructure this so that the debmon repo addition is done inside of the `icinga2::server::install::repos` subclass? An `if`/`else` on [line 59](https://github.com/Icinga/puppet-icinga2/blob/master/manifests/server/install.pp#L59) where if `use_debmon` is `true`, then add it, `else` `include apt::backports`, maybe?

Also, would you be able to rename the parameter itself from `use_debmon` to `use_debmon_repo`? It's kind of a nitpick, but I want to keep the parameter names descriptive.

nickchappell:
Hi @albustax,

Any progress?

albustax:
Hi @nickchappell,

Sorry for the delay, it's done.

nickchappell:
Merged (haven't pushed yet though). Thanks @albustax!

PR #21, opened 2014-09-30
Allow NRPE command argument processing

smbambling:
Adds a parameter to icinga2::nrpe class nrpe_allow_command_argument_processing to allow clients to specify arguemnts to the command line.

By default this is set to 0 ( do not allow arguments) in params.pp


smbambling:
Closing will do a correct request from a feature branch

PR #11, opened 2014-09-17
add icinga2::feature definition

adamcstephens:

PR #23, opened 2014-10-02
Enable/Disable an Icinga2 Service

smbambling:
Similar to Pull #11 

This allows you to specify an array for enabled and disabled features via a parameter under the icinga2::server class.

The features.pp class will check to see if a feature has been added in both arrays and only disable it if it has been.

The enable.pp define will check to make sure the *.conf file exists in features-available before trying to enable it and checks to make sure a file/symlink is created in features-enabled to make sure it does exec multiple times.

Additionally a new object manifest has been added to set the conf file for graphite (GraphiteWriter).


manifests/object/graphitewriter.pp
```diff
@@ -0,0 +1,23 @@
+# Configure GraphiteWriter Feature .conf file
+define icinga2::object::graphitewriter (
+  $host = '127.0.0.1',
```
nickchappell: Could you change these names to `graphite_host` and `graphite_port`? Sorry to nitpick, but I want to keep parameter names descriptive, as the module will get tons of parameters as it grows in size.

nickchappell:
Hi @smbambling,

Would you be able to submit the features-enabled changes and the GraphiteWriter changes as separate PRs?

templates/object_graphitewriter.conf.erb
```diff
@@ -0,0 +1,22 @@
+/**
+ * WARNING: This IdoPgsqlConnection definition is automatically generated by Puppet.
```
nickchappell: Typo here.

manifests/object/graphitewriter.pp
smbambling: Sure thing

smbambling:
Sure I'll split these into two PRs.

manifests/object/graphitewriter.pp
smbambling: Just to note the icinga2::object::idomysqlconnection and icinga2::object::idopgsqlconnection defines have non-descriptive parameters.  Ex. host,port,user ....etc.

nickchappell: Noted. I should have time this weekend to clean most of the smaller stuff up and to test/merge these PRs.

nickchappell:
Hi @smbambling,

I tested this functionality and it looks good, but when I applied the patch, it still brought in the GraphiteWriter commits. Would you be able to rebase and repush this PR against the current master branch with just the commits for the feature enable/disable code?

nickchappell:
Hi @smbambling,

I figured out what was going on and merged this PR in. Thanks!

PR #25, opened 2014-10-02
assign where / ignore where : icinga2 do not start if there are multiple...

MelanieGault:
fixes redmine 7308


nickchappell:
Merged this in. Thanks @MelanieGault!

PR #26, opened 2014-10-03
adding defined type to create a checkcommand object from a source file

smbambling:
Defined type that references a source file for its content as these objects can widely vary between each other.

### DIFF
--- a/manifests/object/checkcommand.pp
+++ b/manifests/object/checkcommand.pp
@@ -1,0 +1,32 @@
+# == Defined type: icinga2::object::checkcommand
+#
+#  This is a defined type for Icinga 2 CheckCommand objects
+# See the following Icinga 2 doc page for more info:
+# http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/configuring-icinga2#objecttype-checkcommand
+
+define icinga2::object::checkcommand (
+  $checkcommand_name              = "${name}.conf",
+  $checkcommand_target_dir        = '/etc/icinga2/objects/checkcommands',
+  $checkcommand_target_file_owner = 'icinga',
+  $checkcommand_target_file_group = 'icinga',
+  $checkcommand_target_file_mode  = '0640',
+  $checkcommand_source_file       = undef,
+) {
+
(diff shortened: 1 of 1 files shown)


smbambling:
Moved from using a source file to a templates.

nickchappell:
Hi @smbambling,

Do you have an ERB template and parameters for the various options in a check command object that you could add?

smbambling:
I currently don't have a template or parameters for a skeleton CheckCommand object.  These can very so much that I'm finding it hard to develop a standard template with parameters.

As in PR #27 I'd make the argument for using templates in place of sourcing a static file with the file server.

nickchappell:
I'm hoping to tackle notify command objects sometime in the next few days, as I think they will have similar sorts of problems. My thoughts on that object type were to use hashes in the Puppet code and then iterators in the ERB template to grab the keys/values from the hashes. Iterating over nested hashes in ERB should be doable. If I can make that work, would you want to refactor this PR modeled after that? If I can make it work the way I have in mind, refactoring this to function in the same way shouldn't be too difficult.

smbambling:
Sure if you wouldn't mind shooting me an example of what your thinking I'll be glad to refactor this request.

cropalato:
i didn't see your pr. I just add a new one PR (29) to fix the same issue.

smbambling:
Your request does seem to follow more of the modules conventions and it nicely done.

I do feel like it's a lot more code and complex then managing a flat file ( served by the file server ) or just managing a template per check command then having all the code in manifest. 

That said this is Nicks module so I'll go with what he decides :)

nickchappell:
I should be able to squeeze in some time to look at #29 and #30 this weekend. @smbambling, after I merge them, would you want to open a PR for each one for adding the ability to specify your own custom template or static file like you did with #27?

At some point in the near future, I may add the ability to do the same thing to the other object types.

smbambling:
Sure I can work on adding that functionality to the object.

After you test a merge in #29 and #30 can you close out this PR ?

smbambling:
Close in favor of request #35

PR #27, opened 2014-10-03
Add independent check plugin defined type

smbambling:
This defined type is independent of NRPE to make it less confusing when using other client that reference the check plugins (scripts) or check plugins (scripts) that are not used via NRPE but executed locally from the Icinga 2 Core node.

Created a new parameter $checkplugin_libdir   = '/usr/lib/nagios/plugins' to keep backwards compatibility with the nrpe::plugins manifest.

### DIFF
--- a/manifests/checkplugin.pp
+++ b/manifests/checkplugin.pp
@@ -9,7 +9,7 @@ define icinga2::checkplugin (
   $checkplugin_target_file_owner  = 'root',
   $checkplugin_target_file_group  = 'root',
   $checkplugin_target_file_mode   = '0755',
-  $checkplugin_source_file        = undef,
+  $checkplugin_template           = undef,
 ) {
 
   #Do some validation of the class' parameters:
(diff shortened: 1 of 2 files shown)


nickchappell:
Hi @smbambling,

Is this for allowing a user to upload their own executable plugin script that the Icinga 2 core would run?

smbambling:
Yes or any client NRPE,NSClient++,ect...

nickchappell:
Would you be able to add support for static files in addition to the template sources?

smbambling:
I think that is possible, with a parameter switch to set which you would like to use source vs content.

Though I'd make the argument for using templates for this case even when having a "static" file when only a single file (excluding binary files)is being managed by a define or class.
- It seems more tidy when not having to mix both static files and templates in a module.  
- Templates are shipped with the catalog and can be used with remote systems that can't access access the Puppet Master/Server file share ( Applied Offline )
- Gives you the flexibility of adding referenced syntax without having to modify the define/class of the module
- Less security management with the puppet file server

There are a few downfalls which IMO are not a concern with this module.
- Templates do put more load on your master for catalog compilation then service static files
- Templates do not support serving 'recursive' File resources

That said if required I'll work on modifying the define to allow using either the file server or templates to serve the file.

nickchappell:
I think the option of using either a template or a static file source would be best.

smbambling:
This has been updated to allow for either using a template or a static file via the file server.

nickchappell:
Merged this in. Thanks @smbambling!

PR #29, opened 2014-10-16
Add a CheckCommand object defined type, refs#7213

cropalato:


nickchappell:

PR #32, opened 2014-10-20
Add a NotificationCommand object defined type, refs#7214

cropalato:


manifests/object/notificationcommand.pp
```diff
@@ -0,0 +1,57 @@
...
+#
+
+define icinga2::object::notificationcommand (
+  $object_notificationcommandname = $name,
+  $template_to_import = 'plugin-check-command',
+/*  $methods           = undef, */ /* Need to get more details about this attribute */
+  $command           = undef,
+  $cmd_path          = 'PluginDir',
+  $arguments         = {},
+  $env               = undef,
```
nickchappell: This will have to be `{}` as a default.

manifests/object/notificationcommand.pp
```diff
@@ -0,0 +1,57 @@
...
+  $vars              = {},
+  $timeout           = undef,
+  $target_dir        = '/etc/icinga2/conf.d',
```
nickchappell: Could this default to `/etc/icinga2/objects/notificationcommands` to match the other object types?

manifests/object/notificationcommand.pp
```diff
@@ -0,0 +1,57 @@
...
+  $target_file_group = 'root',
+  $target_file_mode  = '0644'
+) {
+
+  #Do some validation of the class' parameters:
+  validate_string($object_notificationcommandname)
+  validate_string($template_to_import)
+  validate_array($command)
+  validate_string($cmd_path)
+  if $env {
```
nickchappell: If `$env` defaults to an empty hash, you can drop the `if` conditional.

templates/object_notificationcommand.conf.erb
```diff
@@ -0,0 +1,62 @@
+/**
+* WARNING: This CheckCommand definition is automatically generated by Puppet.
```
nickchappell: Typo.

templates/object_notificationcommand.conf.erb
```diff
@@ -0,0 +1,62 @@
...
+  <%- if @template_to_import -%>
+  import "<%= @template_to_import -%>"
+
+  <%- end -%>
+  <%- if @command -%>
+  command = [ <% if @cmd_path -%><%= @cmd_path -%> + <% end -%><% @command.each_with_index do |cmd, i| %><%= cmd -%><%= ', ' if i < (@command.size - 1) %><% end %> ]
+  <%- end -%>
+  <%#- Need to add support to argument value as hash (recursivity) -%>
+
+  <%- if @arguments -%>
```
nickchappell: Right now, this conditional will allow `arguments = {}` to be placed into the generated object file if `$args` from the Puppet code is an empty hash. `arguments = {}` in the object file will cause errors when Icinga 2 starts.

If the default `args` parameter is an empty `{}` hash, a better conditional check would be:

```
<%- if @arguments.empty? != true -%>
```

This way, we can keep the hash's default value as empty (`{}`), which will still let the `validate_hash()` function work. (`validate_hash` will cause catalog compilation to fail if `$args` is `undef`).

templates/object_notificationcommand.conf.erb
```diff
@@ -0,0 +1,62 @@
...
+      <%- value.each_pair do |k,v| -%>
+      <%= k %> = <%= v %>
+      <%- end -%>
+    }
+    <%- end -%>
+  <%- end -%>
+  }
+  <%- end -%>
+
+  <%- if @vars -%>
```
nickchappell: Same thing as the `<%- if @arguments.empty? != true -%>` thing above.

templates/object_notificationcommand.conf.erb
```diff
@@ -0,0 +1,62 @@
...
+  }
+    <%- end -%>
+  <%- end -%>
+  <%- end -%>
+
+  <%- if @timeout -%>
+  timeout = <%= @timeout %>
+  <%- end -%>
+
+  <%- if @env -%>
```
nickchappell: Same thing as the `<%- if @arguments.empty? != true -%>` thing above.

templates/object_notificationcommand.conf.erb
```diff
@@ -0,0 +1,62 @@
...
+  env = {
+  <%- @vars.each_pair do |key,value| -%>
```
nickchappell: Should this be `@env`?

nickchappell:
Hi @cropalato,

I made some comments inline on the diffs. I test the fixes for them locally, but figured I'd let you make and commit the actual changes so you'd get attribution/credit in the commit logs.

One problem: After the fixes I mentioned, I was able to get Puppet to generate the object file successfully, but Icinga 2 itself wasn't able to parse it. The generated object file didn't have any obvious errors (Gist: https://gist.github.com/nickchappell/5d6ca38806150834db5e). I asked in #Icinga on IRC.

nickchappell:
Update on the error: using the sample code for your object type from the README was creating a duplicate notification command. Renaming it so the object name was different resulted in a different error. I'd hold off on any changes to this PR until I find out more from the Icinga devs, as it may not be an issue with your defined type's code.

manifests/object/notificationcommand.pp
```diff
@@ -0,0 +1,57 @@
...
+# http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/configuring-icinga2#objecttype-notificationcommand
+#
+# === Parameters
+#
+# See the inline comments.
+#
+
+define icinga2::object::notificationcommand (
+  $object_notificationcommandname = $name,
+  $template_to_import = 'plugin-check-command',
```
nickchappell: This should be `plugin-notification-command` instead. Right now, it's trying to load a template for another object type and that causes Icinga 2 to crash for me when it loads.

manifests/object/notificationcommand.pp
cropalato: fixing..

manifests/object/notificationcommand.pp
cropalato: If $env defaults to an empty hash, I need to check if hash is empty inside template, not sure if it is a good idea. What do you think?

templates/object_notificationcommand.conf.erb
cropalato: fixing...

manifests/object/notificationcommand.pp
nickchappell: A check in the template like `<%- if @env.empty? != true -%>` should work. I also suggested the same for `env`.

templates/object_notificationcommand.conf.erb
cropalato: I'm changing arguments default value to undef (manifest). I prefer to keep template simple. what do you think?

templates/object_notificationcommand.conf.erb
cropalato: fixing...

manifests/object/notificationcommand.pp
cropalato: fixing...

cropalato:
@nickchappell, please let me know if you really want to move the variables validation to the template. I fixed the others issues.

nickchappell:
Hi @cropalato,

Yes, if you could set the defaults for `arguments`, `env` and `vars` to empty hashes and move the empty hash check to the template (with just `validate_hash()` in the manifest), that would be good!

cropalato:
Hi @nickchappell,

Done. Please let me know if you can see something else to be fixed.

nickchappell:

PR #33, opened 2014-10-20
Add a EventCommand object defined type, refs#7215

cropalato:


manifests/object/eventcommand.pp
```diff
@@ -0,0 +1,58 @@
...
+  $target_file_name  = "${name}.conf",
+  $target_file_owner = 'root',
+  $target_file_group = 'root',
+  $target_file_mode  = '0644'
+) {
+
+  #Do some validation of the class' parameters:
+  validate_string($object_eventcommandname)
+  validate_string($template_to_import)
+  validate_array($command)
```
nickchappell, line 32: Should we validate this as a string instead? The [official docs](http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/configuring-icinga2#objecttype-eventcommand) say it can be either, but I think it's simpler just to make it a string.

templates/object_eventcommand.conf.erb
```diff
@@ -0,0 +1,62 @@
...
+*/
+
+object EventCommand "<%= @object_eventcommandname %>" {
+  <%#- If any of the @ parameters are undefined, don't print anything for them: -%>
+
+  <%- if @template_to_import -%>
+  import "<%= @template_to_import -%>"
+
+  <%- end -%>
+  <%- if @command -%>
```
nickchappell, line 19: Depending on how we decide to implement the `$command` parameter in `manifests/object/eventcommand.pp`, we might also want to redo this.

templates/object_eventcommand.conf.erb
```diff
@@ -0,0 +1,62 @@
...
+  command = [ <% if @cmd_path -%><%= @cmd_path -%> + <% end -%><% @command.each_with_index do |cmd, i| %><%= cmd -%><%= ', ' if i < (@command.size - 1) %><% end %> ]
+  <%- end -%>
+  <%#- Need to add support to argument value as hash (recursivity) -%>
+
+  <%- if @arguments -%>
```
nickchappell: Hi @cropalato,

Comments on the diffs from #32 for similar variables and syntax things, like `{}` vs `undef` and the different type of `if` conditional would also apply here.

manifests/object/eventcommand.pp
cropalato, line 32: The idea was keep the code as siple an clean as possible. Suport string and
hash will request more code without advantage. do you disagree?
Le 22 oct. 2014 01:09, "Nicholas Chappell" notifications@github.com a
écrit :

cropalato, line 32: sorry. array.
Le 22 oct. 2014 08:26, "Ricardo Cropalato" ricardo@cropalato.com.br a
écrit :

cropalato:
@nickchappell, please let me know if you really want to move the variables validation to the template. I fixed the others issues.

nickchappell:
Hi @cropalato,

Yes, if you could set the defaults for `arguments`, `env` and `vars` to empty hashes and move the empty hash check to the template (with just `validate_hash()` in the manifest), that would be good!

cropalato:
Hi @nickchappell,

Done. Please let me know if you can see something else to be fixed.

nickchappell:

PR #35, opened 2014-10-23
F checkcommand static template

smbambling:
Updated to allow you to configure a check command via a 'plain' file with custom erb template or source file.

Also update the parameter checks to only validate some if using the default .erb template

### DIFF
--- a/manifests/object/checkcommand.pp
+++ b/manifests/object/checkcommand.pp
@@ -11,34 +11,38 @@
 
 define icinga2::object::checkcommand (
   $object_checkcommandname = $name,
-  $template_to_import           = 'plugin-check-command',
-/*  $methods                    = undef, */ /* Need to get more details about this attribute */
-  $command                      = undef,
-  $cmd_path                     = 'PluginDir',
-  $arguments                    = {},
-  $env                          = undef,
-  $vars                         = {},
-  $timeout                      = undef,
-  $target_dir                   = '/etc/icinga2/objects/checkcommands',
-  $checkcommand_template_module = 'icinga2',
-  $checkcommand_template        = 'object_checkcommand.conf.erb',
-  $target_file_name             = "${name}.conf",
-  $target_file_owner            = 'root',
-  $target_file_group            = 'root',
-  $target_file_mode             = '0644'
+  $template_to_import                    = 'plugin-check-command',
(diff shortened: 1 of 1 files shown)


smbambling:
@nickchappell Any suggestions on this PR to get it approved ?

manifests/object/checkcommand.pp
```diff
@@ -11,32 +11,38 @@
...
-  $target_file_name  = "${name}.conf",
-  $target_file_owner = 'root',
-  $target_file_group = 'root',
-  $target_file_mode  = '0644'
+  $template_to_import                    = 'plugin-check-command',
+/*  $methods                             = undef, */ /* Need to get more details about this attribute */
+  $command                               = undef,
+  $cmd_path                              = 'PluginDir',
+  $arguments                             = {},
+  $env                                   = undef,
```
nickchappell, line 19: Would you be able to make this default to an empty hash and change the accompanying code in the ERB template?

The `EventCommand` defined type would be a good model to follow:

https://github.com/Icinga/puppet-icinga2/blob/master/manifests/object/eventcommand.pp

https://github.com/Icinga/puppet-icinga2/blob/master/manifests/object/eventcommand.pp

nickchappell:
Hi @smbambling,

Sorry for letting this sit for so long. Made an inline comment on one of the diffs.

nickchappell:
Just for reference, created a dev.icinga.org issue for this: [https://dev.icinga.org/issues/7628](https://dev.icinga.org/issues/7628)

smbambling:
I'm unclear of the change you are requesting.  The purpose of this PR is to allow you to use an independent template outside the bundled erb provided (object_checkcommand.conf.erb) or allowing you to use a static file to manage the check command.

Could you provide a more detailed change request if required ?

nickchappell:
Hi @smbambling,

Sorry I was being a little too terse last night. I had some back and forth on the other object types about how parameters that are destined to become dictionaries in the final object files should be specified as parameters, how to validate them and how to render the Puppet data in the ERB template.

That's probably a conversation better suited for a separate PR or issue, though. I can merge this in for now and revisit the hash stuff at some point in the future.

nickchappell:
Merged this in. Thanks @smbambling!

PR #39, opened 2014-11-06
Purge unmanaged files from nrpe.d

arioch:
With this patch anything inside /etc/nagios/nrpe.d/ that is not managed by Puppet will be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ For production use, you'll probably want to get the database password via a [Hie
 
 This defined type creates custom files in the `/etc/icinga2/conf.d` directory.
 
-The `icinga2::conf` type has `target_dir`, `target_file_name`, `target_file_owner`, `target_file_group` and `target_file_mode` parameters just like the `icinga2::object` types. 
+The `icinga2::conf` type has `target_dir`, `target_file_name`, `target_file_owner`, `target_file_group` and `target_file_mode` parameters just like the `icinga2::object` types.
 
 The content of the file can be managed with two parameters:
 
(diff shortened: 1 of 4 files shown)


smbambling:
What about also purging items from /etc/nrpe.d just as an extra precaution ?

arioch:
@cropalato: would you like the default to leave the files as-is with the option to enable it? Or the other way around?

@smbambling: good point. I'll add it too.

nickchappell:
@arioch I'd prefer the parameter to default to leaving things as-is. It would be the least surprising behavior. Also, could you add a section to the README explaining the parameter (including default behavior) and showing a usage example?

smbambling:
Agreed the last thing you want to do is have someone unknowingly break their monitoring infrastructure.  Especially if they are transitioning to using the module

arioch:
That's fixed in 22770bf232b6abaeba14602a5e0fb2ff31386df9.

arioch:
Anything else needed to get this merged?

nickchappell:
Hi @arioch,

Would you be able to add the `nrpe_purge_unmanaged` to the parameter list of the `icinga2::nrpe` class and make its default value the value of `nrpe_purge_unmanaged` in `params.pp`? Right now, it can't actually be set.

arioch:
My bad, I was expecting it to be inherited from ::icinga2::params.
Should be fixed now.

nickchappell:
Just for reference, create a [dev.icinga.org](https://dev.icinga.org) issue: [#7673](https://dev.icinga.org/issues/7673)

nickchappell:
Merged. Thanks @arioch, @cropalato and @smbambling!

PR #40, opened 2014-11-07
Fix quoting of PostgreSQL password in exec

jshholland:
This is a little tricky since single quotes cannot be escaped within a single quoted string, but I have now found a solution.


nickchappell:
Hi @jshholland,

Were you having issues with the way the password is currently quoted?

jshholland:
I was, with a (randomly generated) password starting with !% that the shell was trying to perform history expansion on. As far as I can tell this PR fixes it.

nickchappell:
Made a [dev.icinga.org](https://dev.icinga.org) for reference: [7675](https://dev.icinga.org/issues/7675)

nickchappell:
Merged. Thanks @jshholland!

PR #41, opened 2014-11-10
add CentOS5 support

dmizelle:
CentOS 5 is still receiving maintenance updates until 2017, so we should provide support for this case statement using operatingsystemmajrelease.

All the packages in the list are available within EPEL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,10 +118,10 @@ When the `server_db_type` parameter is set, the right IDO database connection pa
 #Install Icinga 2:
 class { 'icinga2::server':
   server_db_type => 'pgsql',
-  db_host => 'localhost'
-  db_port => '5432'
-  db_name => 'icinga2_data'
-  db_user => 'icinga2'
+  db_host => 'localhost',
+  db_port => '5432',
+  db_name => 'icinga2_data',
+  db_user => 'icinga2',
   db_password => hiera('icinga_db_password_key_here'),
 }
 </pre>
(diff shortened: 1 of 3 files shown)


nickchappell:
Hi @dmizelle,

Would you be able to update the README's [Setup](https://github.com/Icinga/puppet-icinga2#setup) section and mention that the NRPE part of the module can be used on EL5?

Would you also be able to update the `metadata.json` file?

dmizelle:
Thanks for the feedback @nickchappell. Hopefully the edits are what you are looking for. Let me know if any additional changes are needed.

nickchappell:
Hi @dmizelle,

Do the server components work on CentOS 5 (don't have a CentOS 5 system handy to test with)?

dmizelle:
Hey @nickchappell . Currently fighting this bug:
https://dev.icinga.org/issues/6807
with shroud in IRC right now. We're trying to see if we can fix it. Server packages won't install until it is resolved.

dmizelle:
@gunnarbeutner just fixed the CentOS 5 repos, so cracking along more on this. Will update this PR when I'm done testing.

dmizelle:
Greetings @nickchappell 

Did some further testing yesterday. Was able to install icinga2 successfully against a postgresql DB on CentOS 5.11 using my changes. Only stipulation is that the user needs PGSQL >9.0, and I have reflected that in the README.

Let me know if there's anything else.

nickchappell:
Just for reference, I created a [dev.icinga.org](https://dev.icinga.org) issue: [#7674](https://dev.icinga.org/issues/7674)

nickchappell:
Merged. Thanks @dmizelle!

PR #42, opened 2014-11-12
Fixed internal links.

cropalato:
@nickchappell these links were not working. It is not a big deal, but as I'm a lazy guys, it was boring me. ;-)


nickchappell:
Merged. Thanks @cropalato!

PR #47, opened 2014-11-13
Adding PerfdataWriter object. #7220

cropalato:


nickchappell:
Just for reference, [#7220](https://dev.icinga.org/issues/7220) is the [dev.icinga.org](https://dev.icinga.org) issue.

nickchappell:
I merged this in. Thanks @cropalato!

PR #50, opened 2014-11-13
Adding ExternalCommandListener object.

cropalato:


cropalato:
https://dev.icinga.org/issues/7224

nickchappell:
I created a [dev.icinga.org](https://dev.icinga.org/) issue for this: [#7224](https://dev.icinga.org/issues/7224)

nickchappell:
Hi @cropalato, the patch for this won't apply because the README has diverged from the state it was in when you made the PR (I've merged a few other PRs tonight). Would you be able to rebase the current master onto your feature branch and re-push?

This page is a good overview on how to do that: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request#how-do-i-rebase

cropalato:
Hi @nickchappell, It should be ok now. As we are changing README.md all the time, how do you think should we avoid this problem. Can I commit without README.md changes and commit README with every 3 or more objects? Or should I continue doing like this and we need to rebase after each commit?
By the way, this week as I'll have ore free time, I'll commit more then usually. =D

templates/object_externalcommandlistener.conf.erb
```diff
@@ -0,0 +1,21 @@
...
+/**
+* A ExternalCommandListener definition. You can create your own configuration files
+* in the conf.d directory (e.g. one per commnand). By default all *.conf
+* files in this directory are included.
+*
+*/
+
+library "compat"
+
+object ExternalCommandListener "<%= @object_externalcommandlistenername %>" {
```
nickchappell, line 15: There's a mismatch between the parameter name here and the one in the manifest, `object_externalcommandlisternername`. This causes the object name in the rendered Icinga 2 file to be empty:

<pre>
object ExternalCommandListener "" {
.....
}
</pre>

cropalato:
Fixed typo.

nickchappell:

PR #51, opened 2014-11-18
Adding hiera support. #7713

cropalato:
The idea is use hiera do define the objects, in this way we can add a data abstraction layer as defined by puppetlabs as good practice.
https://dev.icinga.org/issues/7713


nickchappell:
Hi @cropalato,

Last night, I pushed up a develop branch. To keep the master branch reasonably stable, I'd like to add features and fixes to the develop branch instead of master. Would you be able to resubmit this against the develop branch?

cropalato:
Hi @nickchappell,
moved.

PR #57, opened 2014-11-25
Feature/hash in value vars object host

TheBigLee:
The value of the hash vars can be a hash itself. This should be considered in the erb template.
This should also be done for the other objects.


TheBigLee:
I actually just wanted to add my own commit to the pull request and not 15 other requests.
Can this be fixed?

nickchappell:
Hi @Bigluu,

You'll close this PR and open a new one that's against the current develop branch.

PR #58, opened 2014-11-25
Adding hash support for vars values in object::host

TheBigLee:
Added routine for cases where the value of a vars variable is a hash.


nickchappell:

pdirksen:
Without this patch object::host is pretty much useless because you can't specify the contents of the vars variable right now. Thanks

nickchappell:
Created a [dev.icinga.org](https://dev.icinga.org) issues: [#8156](https://dev.icinga.org/issues/8156)

nickchappell:
Merged this in. Thanks @Bigluu!

PR #59, opened 2014-11-25
Adding CheckerComponent Object. #7227

cropalato:
Hi @nickchappell,
not sure about the directory to save this file. Based on our chat last time I decided move to /etc/icinga2/conf.d. Let me know if you want to move to another place.

https://dev.icinga.org/issues/7227.


manifests/object/checkercomponent.pp
```diff
@@ -0,0 +1,36 @@
...
+# === Parameters
+#
+# See the inline comments.
+#
+
+define icinga2::object::checkercomponent (
+  $ensure                    = 'file',
+  $object_name               = 'checker',
+  $target_dir                = '/etc/icinga2/conf.d',
+  $target_file_name          = "${object_name}.conf",
```
nickchappell: You'll have to set `$target_file_name` to `$name`. The `$name` variable is available and defined when Puppet evaluates the defined type, but it looks like the `$object_name` variable doesn't get 'set' internally until after it tries to create the file resource. When I tested it, it created the file as `/etc/icinga2/conf.d/.conf`.

cropalato:
fixed

nickchappell:

manifests/object/checkercomponent.pp
```diff
@@ -0,0 +1,36 @@
...
+#
+# === Parameters
+#
+# See the inline comments.
+#
+
+define icinga2::object::checkercomponent (
+  $ensure                    = 'file',
+  $object_name               = $name,
+  $target_dir                = '/etc/icinga2/conf.d',
```
nickchappell, line 15: This object type is included with Icinga 2 already. The default object file is in `/etc/icinga2/features-avalable`. Since the feature enable/disable commands look there for files to symlink into `/etc/icinga2/features-enabled`, could you set the default value of the `target_dir` parameter to `/etc/icinga2/features-avaiable`?

README.md
```diff
@@ -538,6 +539,27 @@ icinga2::object::compatlogger { 'daily-log':
...
+Example:
+
+<pre>
+icinga2::object::checkercomponent {'checker':}
+</pre>
+
+This object support the following parameters:
+* `ensure` - Optional parameter used to remove or create the file, Default value is 'file'. Use 'absent' to remove the file.
+* `object_name` - Optional. Used to define file name. default value is 'checker'
+* `target_dir`  - Optional. Define where the conf fil will be created. Default value is '/etc/icinga2/conf.d'
```
nickchappell, line 555: Will want to change this to mention `/etc/icinga2/features-avalable` instead of `conf.d` if you change the default value in the manifest.

lazyfrosch:
Shouldn't we change this to a proper icinga2::feature::checker class instead of a object.

It doesn't make real sense to instantiate it...

nickchappell:
I merged this in. I'll refactor it along with #60 when I refactor the module for the next release.

PR #60, opened 2014-11-25
Adding NotificationComponent object. #7228.

cropalato:
@nickchappell,

Same questions sent with PR #59 

https://dev.icinga.org/issues/7228


manifests/object/notificationcomponent.pp
```diff
@@ -0,0 +1,40 @@
...
+#
+# See the inline comments.
+#
+
+define icinga2::object::notificationcomponent (
+  $ensure                    = 'file',
+  $object_name               = 'notification',
+  $enable_ha                 = undef,
+  $target_dir                = '/etc/icinga2/conf.d',
+  $target_file_name          = "${object_name}.conf",
```
nickchappell: You'll have to set `$target_file_name` to `$name`. The `$name` variable is available and defined when Puppet evaluates the defined type, but it looks like the `$object_name` variable doesn't get 'set' internally until after it tries to create the file resource. When I tested it, it created the file as `/etc/icinga2/conf.d/.conf`.

cropalato:
fixed

nickchappell:

manifests/object/notificationcomponent.pp
```diff
@@ -0,0 +1,40 @@
...
+# === Parameters
+#
+# See the inline comments.
+#
+
+define icinga2::object::notificationcomponent (
+  $ensure                    = 'file',
+  $object_name               = $name,
+  $enable_ha                 = undef,
+  $target_dir                = '/etc/icinga2/conf.d',
```
nickchappell: This object type is included with Icinga 2 already. The default object file is in `/etc/icinga2/features-avalable`. Since the feature enable/disable commands look there for files to symlink into `/etc/icinga2/features-enabled`, could you set the default value of the `target_dir` parameter to `/etc/icinga2/features-avaiable`?

README.md
```diff
@@ -770,6 +771,30 @@ icinga2::object::notificationcommand { 'mail-service-notification':
...
+ 
+<pre> 
+icinga2::object::notificationcomponent {'notification':} 
+</pre> 
+ 
+This object support the following parameters: 
+* `ensure` - Optional parameter used to remove or create the file, Default value is 'file'. Use 'absent' to remove the file. 
+* `object_name` - Optional. Used to define file name. default value is 'checker'
+* `enable_ha` - Optional. Enable the high availability functionality. Only valid in a cluster setup. Default value is true.  
+* `target_dir`  - Optional. Define where the conf fil will be created. Default value is '/etc/icinga2/conf.d'
```
nickchappell: Will want to change this to mention `/etc/icinga2/features-avalable` instead of `conf.d` if you change the default value in the manifest.

lazyfrosch:
Shouldn't we change this to a proper icinga2::feature::notification class instead of a object.

It doesn't make real sense to instantiate it...

nickchappell:
@lazyfrosch.

I think this PR and #59 have same issue: they're for creating object definitions that are already included in Icinga 2, but aren't always enabled.

The IDO DB connection objects are the same way, though they have parameters that have to be set in order to be useful.

For consistency, a refactor of the objects in the module objects that are for things that have to be enabled/disabled would probably be best. Module objects might not be needed for things that are put in features-available by the packages, but if they have any configurable parameters, being able to set those via the module would be nice.

My preference would be to add defined types in the module for the object types in features-available but leave the existing files laid down by the packages there. If a user wants to enable a feature without setting any parameters (either by their choice or because there aren't any to set), they could use the existing feature enable/disable classes the module already has.

If they do want to set them, they could use the defined type to create the object file and then use the feature enable/disable classes to turn it on/off. I would document this as the preferred pattern in the README. 

For now, I can merge both this and #59 as-is and save a refactor of these 2 (and any other objects already in the module that are for features) for the next major version bump, 0.7, since it would be a breaking change.

@cropalato, nothing's wrong with this PR on its own. We just have to figure out where the object type it creates would fit.

lazyfrosch:
A refactoring is really needed here, my idea is:

Deprecate / Link:
- icinga2::server::features
- icinga2::server::features::enable
- icinga2::server::features::disable

And provide:
- icinga2::feature::*

Providing most common features, including the default settings.

We can keep most of the object defined types, but would want to use that code/templates in their corresponding features. And also do not provide any "singleton" feature via object::, like notification, checker, a.s.o.

Thoughts?

nickchappell:
I agree with most of that. I would still like to include defined types for singleton objects, however. 

Some have parameters that can be set, others (like the DB ones) have parameters that have to be set (like DB connection details). 

Even for the singleton objects that don't have parameters, it would still be good to have the ability to manage them via Puppet so inadvertent deletes/edits/moves of the object files by a user don't render their Icinga 2 server non-functional. 

We can include the defined types to manage them in the module but not use them by default. The files laid down by the packages can be the ones symlinked/de-linked by the `icinga2::feature::enable` and `icinga2::feature::disable` classes by default and the files themselves would only be managed by Puppet if a user explicitly used an `icinga2::object::notificationcomponent` object (or one of the other singleton types).

Does that seem sane?

nickchappell:
I merged this in. @lazyfrosch I'll do the things I mentioned in my last post as a part of my refactor (work and life got in the way, which is why I wasn't able to do it sooner).

PR #61, opened 2014-11-26
Adding FileLogger object. #7229.

cropalato:
https://dev.icinga.org/issues/7229


manifests/object/filelogger.pp
```diff
@@ -0,0 +1,45 @@
...
+# See the inline comments.
+#
+
+define icinga2::object::filelogger (
+  $ensure                    = 'file',
+  $object_name               = $name,
+  $path                      = undef,
+  $severity                  = undef,
+  $target_dir                = '/etc/icinga2/conf.d',
+  $target_file_name          = "${object_name}.conf",
```
nickchappell: You'll have to set `$target_file_name` to `$name`. The `$name` variable is available and defined when Puppet evaluates the defined type, but it looks like the `$object_name` variable doesn't get 'set' internally until after it tries to create the file resource. When I tested it, it created the file as `/etc/icinga2/conf.d/.conf`.

cropalato:
fixed

nickchappell:

nickchappell:

PR #62, opened 2014-11-26
Adding IcingaStatusWriter object. #7230

cropalato:
https://dev.icinga.org/issues/7230


manifests/object/icingastatuswriter.pp
```diff
@@ -0,0 +1,44 @@
...
+# http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/configuring-icinga2#objecttype-checkercomponent
+#
+# === Parameters
+#
+# See the inline comments.
+#
+
+define icinga2::object::icingastatuswriter (
+  $ensure                    = 'file',
+  $object_name               = $name,
```
nickchappell, line 14: You'll have to set `$target_file_name` to `$name`. The `$name` variable is available and defined when Puppet evaluates the defined type, but it looks like the `$object_name` variable doesn't get 'set' internally until after it tries to create the file resource. When I tested it, it created the file as `/etc/icinga2/conf.d/.conf`.

cropalato:
Hi @nickchappell, 
I did'n face the same problem than you. But I found a typo and I fixed. thanks.

nickchappell:

nickchappell:

PR #63, opened 2014-11-26
Adding EndPoint object. #7232.

cropalato:
https://dev.icinga.org/issues/7232


manifests/object/endpoint.pp
```diff
@@ -0,0 +1,46 @@
...
+#
+
+define icinga2::object::endpoint (
+  $ensure                    = 'file',
+  $object_name               = $name,
+  $host                      = undef,
+  $port                      = undef,
+  $log_duration              = undef,
+  $target_dir                = '/etc/icinga2/object/endpoints',
+  $target_file_name          = "${object_name}.conf",
```
nickchappell: You'll have to set `$target_file_name` to `$name`. The `$name` variable is available and defined when Puppet evaluates the defined type, but it looks like the `$object_name` variable doesn't get 'set' internally until after it tries to create the file resource. When I tested it, it created the file as `/etc/icinga2/conf.d/.conf`.

smbambling: @nickchappell is correct you can't set variable names from other variables within the same class.

cropalato:
fixed

nickchappell:

nickchappell:

PR #66, opened 2014-12-12
Sorting hashes in erb template and RedHat support

alexxxxx:
This pull request contains two features:
1. Sorting the hashes in erb template to prevent the files to be re-created at every puppet run with different order of the variables
2. Adding RedHat in the list of supported OS.


nickchappell:
Hi @alexxxxx,

Would you be able to split these out into separate PRs?

alexxxxx:
Done, #67 & #68

PR #73, opened 2015-01-12
Adding hiera support. #7713

cropalato:
@nickchappell too much problem squashing the branch. I created a fresh one.


lazyfrosch:
Thanks for your PR, object.pp is a wrong filename, it should be objects.pp

Have you actually tested that code? :)

cropalato:
@lazyfrosch I agree with you, in fact the the directory should be objects too. what do you think @nickchappell ?

nickchappell:
Hi @cropalato, I agree on the filename change. What do you mean the directory?

PR #74, opened 2015-01-13
Remove classes that are empty

lazyfrosch:
We do not need them, there are just a distraction...


lazyfrosch:
PR #73 will actually use icinga2::objects, but it is still the wrong filename

nickchappell:
I merged this in.

PR #77, opened 2015-01-14
Add inline checkplugin file distribution method

arioch:
With this patch we can specify the content of `${checkplugin_libdir}/${checkplugin_name}` inline.

E.g.:

```
::icinga2::checkplugin { 'check_diskstats':
  checkplugin_file_distribution_method => 'inline'
  checkplugin_source_inline            => 'command[check_disks]=/usr/lib64/nagios/plugins/check_disk -w 20 -c 10 -p /'
}
```

Or in my case using hiera-yaml:

```

---
icinga2::checkplugin:
  check_diskstats:
    checkplugin_file_distribution_method: 'inline'
    checkplugin_source_inline:            'command[check_disks]=/usr/lib64/nagios/plugins/check_disk -w 20 -c 10 -p /'
```

PR #78, opened 2015-01-14
Checkcommand source inline

arioch:
Similar to #77, this patch provides a way to provide content for `${target_dir}/${target_file_name}` inline.


lazyfrosch:
I've done something similar in icinga2::config, see PR #79.

nickchappell:
Hi @arioch,

Could you add a note to the README documenting this?

arioch:
I've closed this one myself because it actually duplicated existing functionality.
Not sure what I was thinking when creating this one.

 #77 is still valid.

PR #79, opened 2015-01-14
Refactor and move config handling

lazyfrosch:
There should not be any impact, because icinga2::server::config was not meant
to be called directly.


lazyfrosch:
https://dev.icinga.org/issues/8216

nickchappell:
I actually had this in progress locally (hadn't committed anything) as a part of making a more generic `icinga2::node` class. Would it be OK to hold off on this change until that's complete?

lazyfrosch:
Wouldn't it easier to go step by step?

My change is quite trivial, as it does not affect the way it works, just the code quality :)

lazyfrosch:
I can rebase if you want to merge :)

lazyfrosch:
Merged into develop under ccd9626

PR #84, opened 2015-01-23
Adding hiera support. #7713

cropalato:
Replacing old PR #73.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ Table of Contents
 4. [Usage - How to use the module for various tasks](#usage)
     * [Object type usage](#object_type_usage)
     * [Objects](#objects)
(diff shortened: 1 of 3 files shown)


nickchappell:
Hi @cropalato,

Looks like your README changes didn't make it over. Could you include those?

cropalato:
@nickchappell, 

Added README

nickchappell:
Hi @cropalato,

I merged this in. However, there was a merge conflict with the `class {'icinga2::objects': }` application in `manifests/server.pp` because of some of the changes I made a few days ago to implement the base `::node` class and rework the `::server` to use it. I had to remove the call to `icinga2::objects` to fix it. Would it be OK to just add the calls back into the chained calls that are already there?

PR #85, opened 2015-01-26
Notify nrpe daemon when config changes

arioch:
Quite trivial: when any nrpe related config changes the daemon should be restarted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,7 +213,7 @@ If you would like to install packages to make a `mail` command binary available 
 
 To manage the features that are enabled or disabled on an Icinga 2 server, you can specify them with the `server_enabled_features` and `server_enabled_features` parameters.
 
-The parameters should be given as arrays of single-quoted strings.  
+The parameters should be given as arrays of single-quoted strings.
 
 **Note:** Even if you're only specifying one feature, you will still need to specify it as an array.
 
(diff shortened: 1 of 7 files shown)


nickchappell:
Hi @arioch,

Could you make this a parameter (defaulting to refresh the daemon is fine)?

arioch:
This should do the trick.
I've rebased it into a single commit.

arioch:
I've noticed similar functionality for the Icinga2 daemon was added just recently so I've renamed the nrpe parameter for sake of consistency.

```
-  $nrpe_daemon_notify = true
+  $refresh_nrpe_service = true
```

lazyfrosch:
I don't see a notify connection for `file { "/etc/nagios/nrpe.d/${command_name}.cfg":` anymore, did you miss something while commiting?

lazyfrosch:
Ah I see what you've done, but that would not work, you will have to use `File {}` before declaring the other classes, or it will not work.

Please but that into config.pp and the command.pp, I think thats cleaner.

arioch:
Global resource defaults apply to subclasses as well.
In this case File {} within nrpe will apply to nrpe::command, nrpe::config, nrpe::install, nrpe::plugin and nrpe::service.

Let me know if you still prefer to have it in config.pp and command.pp.

lazyfrosch:
I thought such defaults are required _above_ deeper includes/declarations to work.

Yes I would prefer that, because a user would not look in init.pp for such a definition. IMHO

arioch:
Here you go.

lazyfrosch:
Sorry, found another `$::icinga2::nrpe_daemon_name` must be `$::icinga2::nrpe::nrpe_daemon_name`, or even better `$::icinga2::nrpe::daemon_name`

arioch:
This should do it then.

nickchappell:
I merged this in. Thanks @arioch!

PR #86, opened 2015-01-26
Changing facts to core_facsts

gcmalloc:
Changed to operatingsystemrelease for ubuntu systems. The fact is a core fact now for ubuntu as documented in 
https://docs.puppetlabs.com/facter/1.6/core_facts.html#operatingsystemrelease

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -115,7 +115,7 @@ class icinga2::params {
           $server_plugin_package_install_options = '--no-install-recommends'
         }
         #Fail if we're on any other Ubuntu release:
-        default: { fail("${::operatingsystemmajrelease} is not a supported Ubuntu release version!") }
+        default: { fail("${::operatingsystemrelease} is not a supported Ubuntu release version!") }
       }
     }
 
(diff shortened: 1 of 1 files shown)


nickchappell:
Hi @gcmalloc,

`operatingsystemrelease` returns the same value as `operatingsystemmajrelease` on Ubuntu. Was there a reason you needed `operatingsystemrelease` to be used instead?

gcmalloc:
the fact wasn't available on one of my ubuntu 14.04 as `operatingsystemmajrelease` is only available on facter 1.7 and above.
After some investigation, I discovered that there was two different fact name for the same fact in the current manifest. So I uniformed them to the one specified in the doc for older version of facter.

lazyfrosch:
Should be save for Ubuntu, as they do point releases just in the description:

```
$ cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=12.04
DISTRIB_CODENAME=precise
DISTRIB_DESCRIPTION="Ubuntu 12.04.5 LTS"
```

Debian is different!

nickchappell:
Hi @gcmalloc,

Facter 1.6 hasn't gotten an update since 2013. It's probably been EOL'd.

lazyfrosch:
@gcmalloc thanks for your PR, only problem is that it makes not much sense to only remote `operatingsystemmajrelease` from Ubuntu.

If you'd like to exchange it up everywhere I'll be happy to review a new PR.

gcmalloc:
Ok, thanks for the review anyway.

PR #88, opened 2015-01-28
Fix undefined script variable error

arioch:
Consider following snippet: 

```
apply Service "check_updates" to Host {
  ...
  check_command = "nrpe"
  period = officehours
}
```

which causes:

```
icinga icinga2[26289]: /etc/icinga2/objects/applys/check_updates.conf(20):   period = officehours
icinga icinga2[26289]: ^^^^^^^^^^^
```

... and is fixed by quoting "officehours".


nickchappell:
I merged this in. Thanks @arioch!

PR #89, opened 2015-02-13
Apply Notification objects 'interval' parameter generates invalid configs

FabianPonce:
Fixed 'interval' parameter for the apply notification objects to not be quoted in the generated configuration.


nickchappell:
Merged this in. Thanks @FabianPonce!

PR #69, opened 2014-12-15
Add anchor resources #8077

jshholland:
This ensures that classes are contained properly. See https://dev.icinga.org/issues/8077.

PR #90, opened 2015-02-18
Add acceptance test

mcanevet:


raphink:
:+1:

nickchappell:
Hi @mcanevet,

I tried merging in #69 but it resulted in a dependency cycle when I was also using the object defined types since the object types refresh the Icinga 2 service. Were you able to find a way around that?

Also, this PR was made against the master branch and contains commits from another PR. In the future, could you submit PRs against the develop branch and make sure they contain commits for just one feature or bug fix?

mcanevet:
@nickchappell Could you provide a manifest code that results in a dependency cycle with #69 so that I can reproduce (and hopefully workaround) it?

I'll make my PR towards develop branch in the future

nickchappell:
Hi @mcanevet,

You can use any of the object defined types to create objects to try and repro the dependency cycle. The problem occurred when they would try to refresh the `icinga2` service.

PR #91, opened 2015-02-18
Allow usage without database

mcanevet:


nickchappell:
Hi @mcanevet,

I think the use case of setting up Icinga 2 without a database is already covered by the `icinga2::node` class that's on the develop branch. Does that base class have anything missing that you'd need for your use case?

Also, this PR was made against the master branch and contains commits from another PR. In the future, could you submit PRs against the develop branch and make sure they contain commits for just one feature or bug fix?

mcanevet:
@nickchappell OK I'll work on the develop branch and see if `icinga2::node` fits my needs.

BTW, why don't you release this module on the puppet forge? The community looks there first when looking for a module.

PR #93, opened 2015-03-05
List of icinga2_client_package to be installed can be overiden at the instanciation time

alexxxxx:
List of icinga2_client_package to be installed can be overiden at the instanciation time

Signed-off-by: Alexandre Beche alexandre.beche@gmail.com

### DIFF
--- a/manifests/nrpe.pp
+++ b/manifests/nrpe.pp
@@ -6,6 +6,7 @@
 
 class icinga2::nrpe (
 
+  $icinga2_client_packages                = $icinga2::params::icinga2_client_packages,
   $nrpe_listen_port                       = $icinga2::params::nrpe_listen_port,
   $nrpe_debug_level                       = $icinga2::params::nrpe_debug_level,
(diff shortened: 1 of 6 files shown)


alexxxxx:
Installing all the plugins on the client may be too much when only a few are needed.
Another solution could be to define 2 variables:

$icinga2_client_packages = ['nrpe', 'nagios-plugins-nrpe']
$icinga2_client_plugins = ['nagios-plugins-all', 'nagios-plugins-openmanage', 'nagios-plugins-check-updates']

And only the $icinga2_client_plugins can be overiden.
Any ideas?

Cheers,
Alex

lazyfrosch:
Thanks for your contribution. I've merged your code as e8844e9 into develop, with some smaller adjustments.

Please note that the nrpe class will get some refactoring soon, also some compatibility breaks that will come with a 1.0.0...

Issue: https://dev.icinga.org/issues/8790

PR #95, opened 2015-03-20
Update apt key to full 40characters

exptom:
Latest version of puppetlabs/apt module shows warning on every puppet run if using a short key so this update includes the full 40 character key.


arioch:
Out of curiosity what's the benefit of not using the short notation?

exptom:
The note in the puppetlabs/apt README has this:

"Note: While this module allows the use of short keys, warnings are thrown if a full fingerprint is not used, as they pose a serious security issue by opening you up to collision attacks."

arioch:
Could you rebase that onto 'develop' branch?

PR #96, opened 2015-03-20
Update apt key to full 40characters

exptom:
Sorry got in a mess with the rebase on previous PR #95 so hopefully this is ok?


lazyfrosch:
Thanks for contributing!

PR #97, opened 2015-03-24
Fix typo in group param default value

ckaenzig:


lazyfrosch:
Thanks for noticing, merged to develop

PR #98, opened 2015-03-25
Add param for purging zones.d directory

ckaenzig:
Useful when using automatic zone config propagation and managing
all configs in zones.d/ using puppet.

I'm not sure why you mention the purge_confd parameter being temporary. If you think this zones.d/ purging should be done differently, please let me know. I'm still quite new to Icinga configuration and might have missed something.

### DIFF
--- a/manifests/config.pp
+++ b/manifests/config.pp
@@ -16,7 +16,6 @@ class icinga2::config {
       '/etc/icinga2',
       '/etc/icinga2/pki',
       '/etc/icinga2/scripts',
-      '/etc/icinga2/zones.d',
     ]:
       ensure => directory,
   }
@@ -37,6 +36,7 @@ class icinga2::config {
       '/etc/icinga2/features-available',
       '/etc/icinga2/features-enabled',
       '/etc/icinga2/objects',
+      '/etc/icinga2/zones.d',
     ]:
       ensure  => directory,
       purge   => $::icinga2::purge_configs,


lazyfrosch:
I'm working on some defined types for endpoints and zones.

Could we use `purge_configs` for the zones.d directory?

ckaenzig:
@lazyfrosch Yes, I guess it's ok to use `purge_configs` for that too. I'll update my PR in that way (tomorrow, I hope).

But I don't think I can use the Endpoint and Zone types because I need them to all be configured in one `/etc/icinga2/zones.conf` file. I used `icinga2::conf` with a custom template to build this file. The Host and Service objects related to each zone are then configured in the `zones.d` directory. Using `object::host` and `object::service` types for that works well using the `target_dir` parameter.

lazyfrosch:
What I did till now is to empty zones.conf, and use zone objects in objects/

I'll try to tidy it up and open a WIP PR when I have a bit time

ckaenzig:
Ok, I didn't know that was possible. I might try to do that, I wouldn't need a custom template anymore. Thanks for your feedback!

lazyfrosch:
I have a basic cluster up and running like that, but the zone specific code is in a service module currently

ckaenzig:
There, the only change left in this commit is moving /etc/icinga2/zones.d/ from the list of managed directory to the ones purged when purge_configs is `true`. If you're ok with this, I guess I should squash my commits into a simple one.

lazyfrosch:
Thanks for your addition, merged to develop.

PR #99, opened 2015-04-09
value sollte auch ein Array beinhalten können

bodsch:
values are also an array.
this made it simpler to give more than one argument for check_nrpe


arioch:
Could you adjust the commit message to be in English? Some of us don't speak German.
F.i.: `git commit --amend -m 'Apply service @vars value should be able to include an array'`

lazyfrosch:
There is also #102 that does something similar

arioch:
Closing as #102 and #105 are already merged.

PR #102, opened 2015-04-22
Don't put a vars. array in ""

janvonde:
arrays were written like

```
vars.array = "["foo", "bar"]"
```

which is a wrong syntax. this pull request changes it and removes the extra "" if it is an array. result is:

```
vars.array = ["foo", "bar"]
```


lazyfrosch:
Nice idea, could you please also do that for services?

And it helps us a lot when you open a issue on https://dev.icinga.org/projects/puppet-icinga2

janvonde:
@lazyfrosch sure. see second commit. Next time I will open an issue as well :-)

lazyfrosch:
This should be solved with PR #105 

Finally some safe handling for vars, please test the changes there. :)

PR #103, opened 2015-05-20
Support multiple host templates

mj84:
I wanted to be able to import multiple host templates per host.

My changes basically are:
- change the variable name and type in host.pp
- change object_host.conf.erb to use an array for host templates instead of a string (similar to groups variable)

Regards,
Markus


arioch:
Hi,

Thanks for the contribution.
Could you please rebase this to a single commit and apply it to branch 'develop'?

templates/object_host.conf.erb
```diff
@@ -12,10 +12,12 @@
...
 object Host "<%= @object_hostname %>" {
   <%#- If any of the @ parameters are undefined, don't print anything for them: -%>
-  <%- if @template_to_import -%>
+  <%- if @templates_to_import.length != 0 -%>
   <%#- Otherwise, include the parameter: -%>
-  import "<%= @template_to_import -%>"
+  <%- @templates_to_import.each do |template| -%>
+  import "<%= template -%>"
   <%- end -%>
+  <%- end %>
```
Farzy: Shouldn't this line end with "-%>" instead of "%>" for consistency?

mj84: You're absolutely right, i was just used to having a blank line between my templates and the rest of the conf file :) I just fixed that.

mj84:
@arioch I just rebased my fork.
I think i already chose to apply it to branch 'develop' (mj84 wants to merge 2 commits into Icinga:develop from mj84:master).
Is there something else i need to do?

lazyfrosch:
Thanks for the PR, we should find a clean way for all the templates where "import" is used.

I'd prefer $templates as a parameter, as well as some any2array handling when only a string is handed over.

@mj84 could you do that change? :)

mj84:
@lazyfrosch since puppet does not allow variable reassignment, i'd need to declare a helper variable within icinga2::object::host for the any2array function call.
Also, when we allow the templates to be either a string OR an array, we cannot use any validate_\* function from puppet stdlib.
Furthermore, renaming $templates_to_import to $templates might lead to problems with users who already use this module.

lazyfrosch:
@mj84 I'm aware of that, we are doing a lot of breaking changes in develop right know, so we have a better module for an 1.0.0 release, then we can think about update safety.

lazyfrosch:
Fixed by 4e8930e5989f9078c7ff351cdf6f22495e612236

PR #104, opened 2015-06-01
Add one more level of depth to vars hash

dgoetz:
This allows to add dictionaries to the vars hash to use in apply for rules


lazyfrosch:
Closed in favor of #105

PR #106, opened 2015-06-12
Properly purge unmanaged files in the objects/* subdirectories

optiz0r:
The purge_unmanaged_object_files parameter is applied to the objects
directory with recursion and force flags enabled, however this behaviour
seems to be turned back off by the file resources for each subdirectory
and as such unmanaged files are not really purged.

This commit copies the purge parameters from the parent objects directory
onto each child directory to enable the intended behaviour.

### DIFF
--- a/manifests/server/config.pp
+++ b/manifests/server/config.pp
@@ -25,7 +25,7 @@ class icinga2::server::config inherits icinga2::server {
     $purge_objects = false
     $force_purge = true
   }
-  
+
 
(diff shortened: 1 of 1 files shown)


lazyfrosch:
Thanks for your PR, please note that this module is under heavy refactoring, and currently all changes go into the `develop` branch.

Purging of that directories should already be handled fine by icinga2::config there.

PR #107, opened 2015-06-18
Added  support for Debian 8 (Jessie)

georgemihalcea:
Tested with Debian 7 and Debian 8 and it is working on both.
For Debian 8, packages 'icinga*' from debmon are pinned with priority 500, and default pin priority for debmon repository is 300.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -23,7 +23,7 @@ class icinga2::params {
      #Pick the right package provider:
       $package_provider = 'yum'
     }
-	
+
 	#RedHat systems:
     'RedHat': {
       #Pick the right package provider:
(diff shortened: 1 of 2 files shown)


arioch:
Thanks for the contrib!

Given the fact this is the official Icinga2 module I'd prefer to have the upstream [Icinga repository](http://packages.icinga.org/debian/dists/icinga-jessie/main/binary-amd64/Packages) in the module instead of Debmon's.

georgemihalcea:
You're welcome!

Well, I could change that, if you want me to. I used debmon for Debian Jessie because it was already like this for Debian Wheezy. I could change the repository for both Jessie and Wheezy, I see there are packages for both in Icinga repository. Or should I just add also Icinga repository and also leave debmon repository as an option (for backwards compatibility at least, I assume there are a lot of people that already use this module)?

I'll also try to implement it in the develop branch when I have some time.

arioch:
I've just noticed `if $use_debmon_repo == true {}` in the code.
I think like you said adding upstream next to Debmon as an option is the way to go.

If possible send PR's to develop instead of master.

georgemihalcea:
Sure, I'll change to develop, but it will take me a little more time. There are some changes to do for that and I do not have time right now. 

I'll create the PR when I'm done. 

Thanks for your quick answers!

formorer:
On Thu, 18 Jun 2015, Tom De Vylder wrote:

> Thanks for the contrib!
> 
> Given the fact this is the official Icinga2 module I'd prefer to have the upstream [Icinga repository](http://packages.icinga.org/debian/dists/icinga-jessie/main/binary-amd64/Packages) in the module instead of Debmon's.
> The official one is the debmon one. 

We usually don't recommend using packasges.icinga.org in production.

alex

lazyfrosch:
I agree with preferring debmon as main repo, if someone wants to use latest packages.

What I don't like is the pinning, could you please remove that?

And is there any reason why to duplicate the `apt::source { 'debmon':` ?

georgemihalcea:
I changed the code to remove pinning for Debian 8, but kept if for Debian 7 (because it was already there). To avoid the duplicated `apt::source { 'debmon':` I had to put more conditions in the code, but I think it is ok now. The idea was to have separate code for debian 7 and debian 8, that is what the resource was duplicated before.

I also changed the key for debmon repository to avoid this warning:
`Warning: /Apt_key[Add key: 29D662D2 from Apt::Source debmon]: The id should be a full fingerprint (40 characters), see README.`

I guess this PR could be integrated in `master`, because already Debian 8 is for some time released.
I'll also create another PR for `develop` branch for this.

arioch:
Closing this one as I've merged your other PR (#109). :+1: 
Thanks again!

georgemihalcea:
You're welcome!

PR #108, opened 2015-06-21
Fix bug in OS detection

jbussdieker:


arioch:
Thanks for the contrib!

I've merged only the 'Fix bug in OS detection' part as we already had `icinga2::object::graphitewriter` available.

jbussdieker:
Awesome thank you!!!

PR #110, opened 2015-06-25
Removing double quotes.

cropalato:
I faced problems sometime ago because of some of the double quotes. As it is usually optional, this should be a user decision not something forced. I sent the first version of some of those files and at the beginning was a good idea. Now I'm not 100% sure.

### DIFF
--- a/spec/defines/icinga2_object_host_spec.rb
+++ b/spec/defines/icinga2_object_host_spec.rb
@@ -33,9 +33,9 @@ describe 'icinga2::object::host' do
           :path => '/etc/icinga2/objects/testhost.conf',
           :content => /object Host "testhost"/,
         }) }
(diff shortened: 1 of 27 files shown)


arioch:
Some of the tests seem to explicitly expect double quotes.
Could you adjust those and rebase it into this PR?

cropalato:
@arioch 
Are you talk about travis?

lazyfrosch:
This is invalid config, names and strings have to be quoted... Unless they are config names and variables.

I've added a function lately to properly output any value, and used that for the vars/env/arguments data, see  icinga2_config_value.

This PR does not make any sense, unless someone would provide all the quoting within their parameters, which also does not make sense.

arioch:
Closing in favor of @lazyfrosch's function.

PR #111, opened 2015-07-06
Add Scientific Linux support

joshbeard:
This commit simply adds 'Scientific' to the OS support.  The values appear to be the same as they are for RedHat/CentOS


arioch:
Thanks!
I've merged this upstream, should be in this repo shortly.

PR #112, opened 2015-07-08
Removing /etc/icinga2/features-available from the directory array

cropalato:
There is not reason to remove the files. They are installed by package and do not affect the system. Plus The user won't be able to activate a new feature because the conf file will not be there.


lazyfrosch:
Thanks for your commit, I've moved features-available to the "normal" managed directories.

Merged that to develop

PR #115, opened 2015-07-09
Fix mysql schema load hostname

jkroepke:
Hi!

if the mysql server is not localhost, Exec['mysql_schema_load'] will be always fail.


lazyfrosch:
Thank you, merged to develop

PR #118, opened 2015-07-15
fixed regex of interval in object/notification.pp

moritzfriedrich:
now validates:
- number with more than one digit, e.g. 9,25,128,3600
- number with minute, e.g. 2m, 30m, 1440m

Have I missed one value option, that should be checked?


arioch:
Merged upstream.
Thanks!

PR #120, opened 2015-07-15
Notification template option

moritzfriedrich:
Add is_template option in Notification class.
Are there any other plan to integrate the template or object selection? Else I will implement my way one by one on the other object classes.

The import option was missing in Notification, so I added as its done in other classes.


arioch:
Merged upstream.
Thanks!

PR #121, opened 2015-07-15
Move install classes into their own file

zachfi:
Its undesirable to have class names exist outside of where the
autoloader expects them.

This work simply moves the subclasses in install.pp out into their own
files, namespaced accordingly.


arioch:
Excellent, thanks for catching this blatant error.
Merged upstream.

PR #123, opened 2015-07-24
Add parameter to override the main Icinga2 config template.

Ziaunys:
- This simply adds a parameter to override the default Icinga2 config template to add flexibility.


arioch:
Thanks for the contribution!
Merged upstream.

PR #126, opened 2015-08-05
Make service name order agnostic

roidelapluie:
Puppet versions < 4.0 do not resolve correctly scoped variables in some
cases. It is a lot more order dependent. To avoid that, this commit uses
a service with a static title and a dynamic name => attribute.

For the packages, as the title is an array, it is not possible to use a
name => attribute. To have a static resource I have then chosen to use
an anchor, which is dependent on the packages.

### DIFF
--- a/manifests/nrpe/command.pp
+++ b/manifests/nrpe/command.pp
@@ -21,13 +21,15 @@ define icinga2::nrpe::command (
   validate_string($nrpe_plugin_libdir)
   validate_string($nrpe_plugin_name)
 
+  include icinga2::nrpe
+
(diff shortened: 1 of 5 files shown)


arioch:
Doesn't seem to fix the problem, Julien.
After fixing a syntax error:

```
Finished in 14.32 seconds (files took 0.65254 seconds to load)
189 examples, 21 failures, 4 pending
```

roidelapluie:
189 examples, 0 failures, 4 pending

arioch:
Excellent, thanks.
I owe you a :beer:.

PR #128, opened 2015-08-06
Update target_dir parameter in each object define.

Ziaunys:
- This updates the target_dir parameter in certain object define types
  so they match the path for each object.


arioch:
Excellent!

arioch:
Merged upstream.

PR #130, opened 2015-08-09
Work feature8297

Reamer:
Hi,

i want help you with Module Version 1.0.
I take a look at this redmine ticket https://dev.icinga.org/issues/8297

Is this a direction which you desired?

In general: I think we should use more includes then direct class calls

Greetings Reamer


lazyfrosch:
Thanks for your addition, merged in, but squashed as one commit

PR #131, opened 2015-08-10
(fix): Add TicketSalt param to ApiListener Object

jrodriguezjr:


lazyfrosch:
While you are right to add that missing parameter, we really want to recommend using the Puppet certificates for Icinga 2.

TicketSalt is basically a private password, that should only be used once, when getting the certificate, and not being stored on any system.

The Puppet way of requesting and signing is already there, and can easily be re-used, by ensuring the certificate files to /etc/icinga2/...

jrodriguezjr:
That is fine but should that not be optional and at the desecration of the implementor?

lazyfrosch:
Yeah we couldn't even stop the user i suppose :)

Thank you, merged.

PR #132, opened 2015-08-12
Fixing require path

bountin:
I've experienced the issue that the generation of host objects with vars does not work but produces an error that a file is not found (For more details see http://stackoverflow.com/questions/31973553/puppet-template-does-not-find-module-function) 

After some time debugging and searching around I found this [blogpost](http://jcg.wtf/blog/2011/03/calling-custom-functions-from-other-custom-functions-in-puppet/) and another similar [PR](https://github.com/jenkinsci/puppet-jenkins/pull/347/files).

I don't really know if this is the correct way to fix this issue but it works in my environment (puppet 4.2.1 with a puppetserver)


lazyfrosch:
Very interesting, doesn't happen in my environments.

Merged. Thank you!

PR #133, opened 2015-08-14
Graphite object 

Reamer:
Hi,

Graphite object should be an unique feature, but we should give the possibility to add more then one object.
At the moment i don't know if this is ever possibly.

refs #8786


lazyfrosch:
Thank you, merged

PR #134, opened 2015-08-14
Update more object define types to be consistent.

Ziaunys:
- This is a follow up to a PR I made a couple weeks ago which just makes sure the other define types default to being created under the appropriate directories.
- It also updates the tests for the ido objects and changes the titles of some file resources so they're consistent with the other define types.

### DIFF
--- a/manifests/feature/ido_mysql.pp
+++ b/manifests/feature/ido_mysql.pp
@@ -51,7 +51,8 @@ class icinga2::feature::ido_mysql (
     enable_ha            => $enable_ha,
     cleanup              => $cleanup,
     categories           => $categories,
-    target_file          => '/etc/icinga2/features-available/ido-mysql.conf',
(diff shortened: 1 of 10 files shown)


Reamer:
:+1:

Reamer:
Hi Ziaunys,
i have tested your patch and there are some mistakes.

The right path should be:
'/etc/icinga2/objects/idomysqlconnection**s**'
'/etc/icinga2/objects/idopgsqlconnection**s**'
'/etc/icinga2/objects/service**s**'

This folders are created in this file: https://github.com/Icinga/puppet-icinga2/blob/develop/manifests/config.pp#L51

Greetings Reamer

Ziaunys:
Thanks for catching that. I went and fixed those paths.

arioch:
Excellent!
Merged upstream.

PR #137, opened 2015-09-03
Allow nrpe commands to be executed with sudo

TheBigLee:
Some NRPE commands require sudo rights. This pull request introduces a new paramter $use_use. If it is set to true it will prepend sudo to the command to run and will also generate a sudoers file for that specific command.


arioch:
Hi,

Looks good so far.
However could you add a little example to readme.md and add some spec tests?

TheBigLee:
I have added a few general examples for the nrpe::command class as it was missing altogether. I have also added an example for the use_sudo parameter.

About the spec tests: The current master branch does not have any spec tests defined.
I do not have the time to implement the whole spec tests from scratch for this branch and is, in my opinion, of scope of this PR.

arioch:
Oh I hadn't noticed you branched from master.
Could you rebase these commits onto develop? There's already a bunch of tests available there.

arioch:
Same goes for your other PR's.

TheBigLee:
I actually intended to do it on the master branch as we still use the 0.x versions and it doesn't look like 1.x will be ready for production use in the near future. If I find the time to write spec tests and adopt this feature for the new 1.x versions I'm happy to do so.

lazyfrosch:
Please move the NRPE code into another module, and use that.

I will deprecate that part with release 1.0.0

PR #138, opened 2015-09-03
Add option to check config before icinga restart

TheBigLee:
If the configuration contains errors the icinga service will not restart
any more. The option configtest_enable allows to first test the
configuration and only restart the service if the configuration does not
contain any errors.
This prevents icinga from not starting anymore.

### DIFF
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -14,7 +14,7 @@
 class icinga2::server::service (
   $configtest_enable = $::icinga2::server::configtest_enable,
   $service_restart   = $::icinga2::server::service_restart,
-) inherits icinga2::server {
+){
 
   include icinga2::server
 
--- a/manifests/server/service.pp
+++ b/manifests/server/service.pp
@@ -11,7 +11,10 @@
 # Coming soon...
 #
 
-class icinga2::server::service inherits icinga2::server {
+class icinga2::server::service (
+  $configtest_enable = $::icinga2::server::configtest_enable,
+  $service_restart   = $::icinga2::server::service_restart,
+) inherits icinga2::server {
 
   include icinga2::server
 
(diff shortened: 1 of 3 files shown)


manifests/server/service.pp
```diff
@@ -11,7 +11,10 @@
 # Coming soon...
 #
 
-class icinga2::server::service inherits icinga2::server {
+class icinga2::server::service (
+  $configtest_enable = $::icinga2::server::configtest_enable,
+  $service_restart   = $::icinga2::server::service_restart,
+) inherits icinga2::server {
```
arioch: Inheritance is deprecated in recent puppet versions...

TheBigLee:
Thanks for the info. I did not know that this is deprecated. I have removed the inheritance from the service class

jkroepke:
What about https://docs.puppetlabs.com/references/latest/type.html#file-attribute-validate_cmd ?

TheBigLee:
I don't think that this would be a good choice in this particular case as this attribute would have to be set for every file definition in every class in the manifests/objects directory.

But this is not my decision to make. If the maintainer of this puppet module agrees with you than I'm happy to go with the validate_cmd attribute.

Reamer:
Why you doesn't use the service script (on CentOS 7 - /sbin/service). With service you are able to support more systems

I agree with @Bigluu validate_cmd in every file resource is not good +
validate_cmd is only available with >= puppet 3.5
https://docs.puppetlabs.com/puppet/3.5/reference/release_notes.html

In general I would prefer that icinga2 is stopped with a wrong configuration. Then we have no mismatch between running config and config on disk.

arioch:
Personally I'd prefer not to move away from the [service](http://docs.puppetlabs.com/references/4.2.latest/type.html#service) resource type.
It's there to avoid confusion and having to implement additional workarounds to support every possible init system.

While this PR does raise a valid point I think we should put more time and thought in this before merging.

lazyfrosch:
@arioch will push a PR with a nicer solution, that works very well

lazyfrosch:
Feel free to comment on my PR :)

arioch:
Closing in favor of @lazyfrosch's solution: #151. 
It's more elegant and uses the native resource type.

PR #139, opened 2015-09-03
Adding hash support for vars values in object::apply_service_to_host

TheBigLee:
The value of the vars variable can be a hash itself. Therefore we need to check for a hash and iterate over that hash and print the subkeys and subvalues. 

This PR is similar to https://github.com/Icinga/puppet-icinga2/pull/58


arioch:
Hi,

Looks good so far.
However could you add a little example to readme.md and add some spec tests?

TheBigLee:
I will add an example later on.
As I have mentioned in another PR, there are currently no spec tests implemented on the master branch.

lazyfrosch:
Sorry for abandoning that PR so long, this should have been implemented with `icinga2_config_value`

PR #140, opened 2015-09-17
Add zone parameter to service object.

Ziaunys:
- The Icinga2 service object has a 'zone' parameter so this updates the Puppet resource to be consistent with the actual object.


Reamer:
works very well as expected
:+1:

arioch:
Merged upstream.
Thanks! :+1:

PR #141, opened 2015-09-28
Add zone parameter to host object.

Ziaunys:
- This updates the host object with a new 'zone' parameter which makes it more consistent with the Icinga2 Host object.


arioch:
Merged upstream.
Thanks! :+1:

PR #144, opened 2015-10-04
opentsdb feature

Reamer:
Hi,

that's a default disabled feature after rpm installation. The opentsdb implementation is a class , so we have maximum one interface. At the moment i see no need for more.

 If the maintainer of this puppet module prefer a define, i can write it.

Greetings Reamer

### DIFF
--- a/README.md
+++ b/README.md
@@ -989,6 +989,23 @@ icinga2::object::graphitewriter { 'graphite_relay':
 }
 ````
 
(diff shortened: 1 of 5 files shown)


lazyfrosch:
Don't want to nag, but a defined type is prefered for this as well ;)

Like #146 
Thanks for your contributions!

Reamer:
Have done the same rewrite like in #146 

Greetings Reamer

lazyfrosch:
Merged, thanks!

PR #145, opened 2015-10-04
Debug Feature

Reamer:
Hi,

that's a default disabled feature after rpm installation. I know that the debug feature is nearly the same like the mainlog feature.

If the maintainer of this puppet module prefer a define for object FileLogger, i can write it.

Greetings Reamer

### DIFF
--- a/templates/feature/debuglog.conf.erb
+++ b/templates/feature/debuglog.conf.erb
@@ -1,0 +1,6 @@
+<%= ERB.new(File.read(File.expand_path('_header.erb',File.dirname(File.dirname(file))))).result(binding) -%>
+
+object FileLogger "debug-file" {
+  severity = "<%= @severity %>"
+  path = "<%= @path %>"
+}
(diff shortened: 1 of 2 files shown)


lazyfrosch:
Thanks!

templates/feature/debuglog.conf.erb
```diff
@@ -0,0 +1,6 @@
+<%= ERB.new(File.read(File.expand_path('_header.erb',File.dirname(File.dirname(file))))).result(binding) -%>
```
zachfi, line 1: What is this line even doing?

Reamer, line 1: Generates a String from this template:
https://github.com/Icinga/puppet-icinga2/blob/develop/templates/_header.erb

Now every File has same header.

zachfi, line 1: There is a better pattern if you want to include a template in another template.  See https://github.com/xaque208/puppet-bacula/blob/master/templates/bacula-sd-header.erb#L11

This effectively renders the template `bacula/_sslkeypair.erb` with erb.

zachfi, line 1: https://github.com/Icinga/puppet-icinga2/pull/154

PR #147, opened 2015-10-15
Deprecate icinga2::server

zachfi:
This work removes references to the icinga2::server class.  This has not
been vetted to be sure that documentation is now correct, but I will be
working through standing up a server without this class and expect to be
making documentation and test updates as I go.


zachfi:
Closes https://dev.icinga.org/issues/8800

lazyfrosch:
Yeah lets actually remove that class!

Thanks.

PR #149, opened 2015-10-20
Add note on using categories

tarjei:
Categories should not be set :)

15:20
(kl. 15.22 +0200) shroud: out of curiosity
(kl. 15.23 +0200) shroud: does the puppet module not set DbCatProgramStatus by default or is that something you changed yourself?
(kl. 15.24 +0200) tarjei: the example sets it:     https://github.com/Icinga/puppet-icinga2/blob/develop/README.md
(kl. 15.24 +0200) shroud: huh
(kl. 15.24 +0200) shroud: that's just silly :P
(kl. 15.24 +0200) shroud: submit a bug report for that
(kl. 15.24 +0200) shroud: because, well, that's just not gonna work with icingaweb2

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,7 @@ icinga2::object::idopgsqlconnection { 'postgres_connection':
    user             => 'icinga2',
    password         => 'password',
    database         => 'icinga2_data',
+   # you should not set categories if you do not have special needs!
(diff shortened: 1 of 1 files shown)


lazyfrosch:
Thanks, I will actually remove this from the documentation...

PR #150, opened 2015-10-21
Remove NRPE from the icinga2 module

lazyfrosch:
fixes #10416

Can I get any thoughts for that?


arioch:
Haven't tested myself but looks good to me!
Finally someone had the courage to remove the NRPE cruft. :+1: 

Are you going to split it to a separate module or just dropping it altogether?

lazyfrosch:
@arioch I'd leave the NRPE module for the community, moving the manifests they already use couldn't be that hard.

I've no current use in my environments.

arioch:
:+1:

lazyfrosch:
Let's do it

PR #151, opened 2015-10-21
Implement config test and stamping for the service

lazyfrosch:
The icinga2 service will not be restarted when the config check failed.

Also a config stamp is written, that will with retrying a restart.

Also: Change notifications to Service class, not the resource

closes #138
refs #10424

### DIFF
--- a/manifests/conf.pp
+++ b/manifests/conf.pp
@@ -50,7 +50,7 @@ define icinga2::conf (
     owner   => $target_file_owner,
     group   => $target_file_group,
     mode    => $target_file_mode,
-    notify  => Service['icinga2'],
+    notify  => Class['::icinga2::service'],
     source  => $manage_file_source,
     content => $manage_file_content,
   }
--- a/manifests/feature.pp
+++ b/manifests/feature.pp
@@ -47,7 +47,7 @@ define icinga2::feature (
   }
 
   if $::icinga2::manage_service {
-    File["icinga2 feature ${name} enabled"] ~> Class['icinga2::service']
+    File["icinga2 feature ${name} enabled"] ~> Class['::icinga2::service']
   }
 
 }
--- a/manifests/feature/api.pp
+++ b/manifests/feature/api.pp
@@ -42,7 +42,7 @@ class icinga2::feature::api (
   }
 
   if $::icinga2::manage_service {
-    $notify = Class['icinga2::service']
+    $notify = Class['::icinga2::service']
   }
   else {
     $notify = undef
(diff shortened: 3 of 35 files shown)


lazyfrosch:
@arioch please have a look

arioch:
Looks good to me. :+1: 
While you're at it it might be a good thing to adjust the class lookup path: 

from `notify  => Class['icinga2::service'],` 
to `notify  => Class['::icinga2::service'],`

Reamer:
Using the notification to Service class is very good.
Why you need the extra config-stamp file?

I think this should also work.
There are three possibilities:
1) config change to valid config
config change -> notify to service class -> check config -> notify service resource -> restart icinga 
2) config change to invalid config
config change -> notify to service class -> check config -> service resource failed because of dependency -> no restart of icinga
3) no config change
service resource will check if icinga is running

```
  exec { 'icinga2 daemon config test':
    command => 'icinga2 daemon -C',
    refreshonly  => true,
  } ~>

  service { 'icinga2':
    ensure     => running,
    enable     => true,
    hasrestart => true,
  }
```

Feedback is welcome. I made no tests

lazyfrosch:
@arioch are you sure this is still recommended? I heard Puppetlabs wants to deprecate the relative class lookups. My notation should be safe either way

@Reamer the config stamp will make sure that you are not left with a onetime error message, while the 
daemon is running an old config, it will try to check/reload the service, until the config is fixed

arioch:
@lazyfrosch `::icinga::service` is static not dynamic (relative). 
Dynamic lookup is removed in Puppet 4.

Reamer:
@lazyfrosch good point. Thank you for make this clear

lazyfrosch:
@arioch I get that, but when there is no dynamic lookup in the future, there should not be a problem leaving the trailing `::` out, or?

arioch:
Leading, not trailing. :wink: 
Leaving it out makes it dynamic. Hence Puppet will look for it in its current context. 
This could f.i. be `::icinga2` but it could also be `::profile::monitoring::icinga2` or anything else calling the module rather than go back to top level (`::`) and walk the given static path down to the right location.

arioch:
Slightly more elaborate:

```
class profile::monitoring {
  include icinga2
}

class profile::monitoring::icinga2 {
  notice 'bad practice'
}
```

When applying the snippet above you'd get to see 'bad practice' returned to stdout. This could potentially lead to all kinds of unexpected behaviour. 

However with the following snippet you'd actually include the icinga2 module as expected because we force Puppet to start looking from the highest level instead of down the subclass path first.

```
class profile::monitoring {
  include ::icinga2
}

class profile::monitoring::icinga2 {
  notice 'bad practice'
}
```

Reamer:
@arioch  :+1:

lazyfrosch:
Merging to develop...

PR #153, opened 2015-10-23
Rework API / Zone / Endpoint features

lazyfrosch:
Not completely tested, but I did those implementation for a user of the module base.

This was tested with a simplified version of the icinga2 module.


lazyfrosch:
Especially specs tests are pending :)

Reamer:
I like the new define "icinga2::config::objectdir"

Why you delete the param "object_name"? I'm using it for exported resources, because zone name must be equal on icinga2 client and icinga2 master. Without objectname i have a duplicated key in catalog. Do you have an other idea how i can manage zones and endpoints automatically with puppet?

Greetings Reamer

lazyfrosch:
@Reamer thanks for asking, I'm working on making that work out of the box. Still to come.

I want to remove object_name, because there is already name. And such tricks you do now should not be needed later.

Reamer:
@lazyfrosch "Out of the box" sounds amazing. What is your basic idea?

lazyfrosch:
@Reamer well, one can do a lot with exported resources, the new implementation of zone, which generated the respective endpoints and lists them as members helps a lot.

When I have time I'll continue this and put examples in the README

lazyfrosch:
I cleaned up the commit history on this.

I will merge this today, so any last thoughts @arioch :question: :smiley:

lazyfrosch:
@Reamer docs are still to be updated.

Since you were curious:

``` puppet
  class { '::icinga2::feature::api':
    accept_commands => true,
    accept_config   => true,
    manage_zone     => false,  # master = false, agent = true
    #parent_zone => 'master',  # only on agents
  }
  include ::icinga2::pki::puppet
```

The master zone has to be specified:

``` puppet
icinga2::object::zone { 'master':
  endpoints => {
    'icinga2a.icinga.local': => {
      'host' => '1.1.1.1',
    },
    'icinga2b.icinga.local': => {
      'host' => '1.1.1.2',
    },
  },
}
```

lazyfrosch:
The last commits allows the zones being defined via hiera:

``` puppet
include ::icinga2::objects
```

``` yaml
icinga2::object::zone:
  master:
    endpoints:
      icinga-testing1.icinga.local:
        host: 'icinga-testing1.icinga.local'
      icinga-testing2.icinga.local:
        host: 'icinga-testing2.icinga.local'
```

arioch:
@lazyfrosch I've taken a quick look and it looks a lot cleaner than it used to be.
If you want I can do a more in depth review later this evening but I trust your judgement so feel free to merge. :+1: 

I'd like to get this module Puppetlabs approved at some point in time. Sadly it's not allowed by Puppetlabs to have hiera_*() calls in a module.

Using autoloading we basically have the same functionality available.

```
if $hash_zone {
  create_resources(icinga2::object::zone,$hash_zone)
}
```

arioch:
hiera

lazyfrosch:
@arioch a basic look is all I wanted.

There is a lot of hiera_\* code in `icinga2::objects`, but it is not included by default.

I guess we would have to rename it to `icinga2::object`, so all the hashes get autoloaded into that class.

lazyfrosch:
Last but not least some spec and compatibility issues on older versions of Ruby and Puppet fixed

lazyfrosch:
Okay, all tests are green.

Please see https://dev.icinga.org/issues/10802 for the hiera stuff

arioch:
:+1:

PR #154, opened 2015-10-26
Update template includes for readability

zachfi:
This work simply replaces the in line ruby calls to the ERB module with
puppet function calls to render the same template.


Reamer:
Looks much better. :+1:

lazyfrosch:
@xaque208 thanks, merged!

PR #156, opened 2015-10-29
Travis updates

lazyfrosch:
This will do 3 things:
- Fix icinga2_config_value on Ruby 1.8.7
- Allow rspec tests on Ruby 1.8 with an older version of rspec
- Update travis definitions with proper values

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,8 +14,7 @@ bundler_args: --without system_tests
 
 matrix:
   allow_failures:
-    - env: PUPPET_VERSION="3.7" FUTURE_PARSER="yes"
-    - env: PUPPET_VERSION="3.8" FUTURE_PARSER="yes"
+    - rvm: 1.8.7
   exclude:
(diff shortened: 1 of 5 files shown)


lazyfrosch:
@arioch can I get your opinion on this?

Will cleanup the commits prior to merge

.travis.yml
```diff
@@ -3,14 +3,36 @@ language: ruby
 sudo: false
 
 rvm:
+  - 1.8.7
```
zachfi, line 6: Why oh why?

lazyfrosch, line 6: Well, because RHEL5 is still a widely used system, and not everyone is using AIO or PE

Reamer, line 6: RHEL6 use ruby 1.8.7, too

arioch:
Looks good to me. :+1:

lazyfrosch:
Merged

PR #146, opened 2015-10-04
gelf feature

Reamer:
Hi,

that's a default disabled feature after rpm installation. The gelf implementation is a class , so we have maximum one interface to an gelf service. At the moment i see no need for more.

If the maintainer of this puppet module prefer a define, i can write it.

Greetings Reamer

PR #157, opened 2015-10-29
Add graphite test + rewrite graphite object + docu

Reamer:
Hi,

some little changes for graphite object, changes in docu after last merge and tests for graphite

Greetings Reamer

### DIFF
--- a/manifests/object/graphitewriter.pp
+++ b/manifests/object/graphitewriter.pp
@@ -10,48 +10,25 @@
 #
 
 define icinga2::object::graphitewriter (
-  $graphite_host        = '127.0.0.1',
-  $graphite_port        = 2003,
+  $host       = '127.0.0.1',
+  $port       = 2003,
   #Put the object files this defined type generates in features-available
   #since the Graphite writer feature is one that has to be explicitly enabled.
-  $target_dir           = '/etc/icinga2/features-available',
-  $target_file_name     = "${name}.conf",
-  $target_file_ensure   = file,
-  $target_file_owner    = 'root',
-  $target_file_group    = 'root',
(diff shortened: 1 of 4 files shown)


manifests/object/graphitewriter.pp
```diff
@@ -10,48 +10,25 @@
 #
 
 define icinga2::object::graphitewriter (
-  $graphite_host        = '127.0.0.1',
-  $graphite_port        = 2003,
+  $host       = '127.0.0.1',
+  $port       = 2003,
```
arioch, line 14: I fear $host and $port are or may to end up on puppet's reserved keywords list.
So it's safer to keep the prefix.

Reamer:
@arioch In PR https://github.com/Icinga/puppet-icinga2/pull/146 @lazyfrosch wrote "avoid the gelf_prefix"
So whats the best way? I don't think that $host and $port may to end up on puppet's reserved keywords list.
$port is used in a lot of other module (example: https://github.com/puppetlabs/puppetlabs-apache/blob/master/manifests/vhost.pp)
$host is also used in popular other module (example: https://github.com/puppetlabs/puppetlabs-mysql/blob/master/manifests/db.pp)

In my own modules i also avoid prefix, because is a double in hiera icinga2::object::**graphite**writer::**graphite**_host

Greetings Reamer

lazyfrosch:
I'm with dropping that prefix. We can't predict everything.

And @Reamer's examples proves a good practice, even from Puppetlabs.

arioch:
Alright, guess you guys have a point.
Merged and thanks again, @Reamer keep those patches coming. :+1:

PR #158, opened 2015-11-01
Added the possibility to reload Icinga2 when the service receives a notification

Reamer:
Hi,

i think we should give the possibility to use the advantages of reloading icinga2.

Greetings Reamer

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,28 @@ class { 'icinga2':
 }
 ````
 
+**Switch from restart to reload Icinga2 service**
+
+The benefits to reload the service can be found in this [chapter](http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/migration#differences-1x-2-real-reload).
+
+Because the Puppet [Resource Service](https://docs.puppetlabs.com/references/latest/type.html#service) doesn't support a reload, you can override the restart command. Every time the service should be restart, we execute a reload.
+
+````
+class { 'icinga2':
+  ...
+  restart_cmd  => 'service icinga2 reload',
+  ...
+}
+````
+
+You can also use Hiera
+
+````
+icinga2::restart_cmd: 'service icinga2 reload'
+````
+
+You should validate the reload command for your operatingsystem.
+
 ### Check Plugins
 
 Agents installed on nodes that Icinga is performing active checks against often require additional or custom check plugins. In order to deploy these check pluings on a node you can call the checkplugin defined resource.
--- a/manifests/init.pp
+++ b/manifests/init.pp
@@ -37,6 +37,7 @@ class icinga2 (
   $db_schema_mysql                        = $::icinga2::params::db_schema_mysql,
   $db_schema_pgsql                        = $::icinga2::params::db_schema_pgsql,
   $pid_file                               = $::icinga2::params::pid_file,
+  $restart_cmd                            = $::icinga2::params::restart_cmd,
 ) inherits ::icinga2::params {
   # TODO: temporary parameter until we provide some default templates
   validate_bool($purge_confd)
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -39,6 +39,8 @@ class icinga2::params {
 
   $pid_file = '/run/icinga2/icinga2.pid'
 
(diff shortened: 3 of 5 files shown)


lazyfrosch:
Thanks for your PR.

Basically this is a good idea, I see 2 problems:
- there is no real good way of reloading a Puppet service, that works across platforms
- Icinga2 internally does a restart every time

There is no real sense to signal a reload.

What was important to me is that we check the config internally before triggering a restart.

The only benefit I can see is that the reload is a bit faster internally, because Icinga2 makes the hand over.

@arioch @dnsmichi your thoughts?

arioch:
We've discussed this before in an earlier ticket. Can't seem to find which one exactly.
However @lazyfrosch has summarised the gist of it. Especially the lack of cross platform support is troublesome.

I've had setups with +3000 line config files reload in a matter of seconds.
I don't see any added value in a reload over a restart.

dnsmichi:
Icinga2 does not do a restart when a SIGHUP is sent. It instead forks a child process which validates the configuration and signals the parent process when it is ready to take over. In the meantime the parent process continues to execute checks, send/receive cluster events, etc. 

When the child wants to take over uppn successful config load, the parent process shuts down, writes its state file, and the child reads its history. That way the downtime is not measurable, and helps a lot in large scale systems.

The old fashioned "restart" is just a plain dumb SIGKILL and the start the daemon again. Which means that the config validation & compiling could take longer (some instances with 100k objects will take e.g. a couple of minutes to do just that). The downtime window being unable to do checks or communicate in the cluster is quite longer.

So the feature request to issue a reload is totally valid. Though I don't know how to implement that cross-/multi-platform with Puppet.

Reamer:
The lack of cross platform support is not the task of this module,it's a task for puppet. But I think we should give the opportunity to people, who would like to reload the icinga2 instance. When i read this [documentation](http://docs.icinga.org/icinga2/latest/doc/module/icinga2/chapter/migration#differences-1x-2-real-reload) it sounds that some people have problem with a normal restart.

I'm new in icingaweb2 dashboard, but i see some interesting stuff. You can add comments with no persistence until icinga2 is restarted. I test this functionality with reload, restart and stop/start but the comment is still there. Maybe sometimes it works. A restart would delete this comment, but a reload hopefully not. Especially when we add/remove dynamical configuration with puppet a restart is not helpful.

arioch:
I accidentally closed this discussion, I meant to just comment. 
My apologies for that.

lazyfrosch:
@dnsmichi agreed, but there is no clean way to do it with Puppet.

Usually a restart is enough, and the config is validated before to make sure it works.

The problem remains is the time gap of shutdown and config test to bring up the new daemon. In the Puppet world, all software managed by Puppet has that problem, e.g. Apache.

If someone comes up with a good solution, I'd consider it, `restart_cmd => 'service icinga2 reload'` is not enough, sorry.

Reamer:
@dnsmichi 
I know there is no clean way. There are several puppet module, with workarounds. The best one is [puppet/postgresql](https://github.com/puppetlabs/puppetlabs-postgresql/blob/master/manifests/server/config_entry.pp)

They default behaviour is still the normal restart, but for big icinga2 instance (master Server) we should give the possibility to reload the service.
I know it's not a good way to override the restart_cmd but it's better then nothing.

A other example is this module [mayflower/php](https://github.com/mayflower/puppet-php/blob/master/manifests/fpm/service.pp). Here is the default behaviour to reload the service.

Greetings Reamer

lazyfrosch:
Okay, I'll give it a try in a test environment. I guess `service` should be working on all the distros we want to support.

Regarding config test, Puppet should actually be able to see that the config is newer than the daemon, since the reload won't update the PID file when the reload failed for whatever reason.

Reamer:
rebase of commit, because PR is two month old.

lazyfrosch:
Am I wrong or have you changed it so there is no default reload_cmd set?

Reamer:
there is a default in params.pp 

```
$restart_cmd = undef
```

lazyfrosch:
Merged, thanks for your contribution

PR #159, opened 2015-11-03
Add command_endpoint for host objects

hashjo:
Added command_endpoint to the host object. This allows more easier configuration for nodealive check to the host in a distributed setup.  This can be used as a var to reference for service_apply command_endpoint. I have tested this configuration in my test setup and it works great. Icingaweb2 even reports the correct Check Source.

### DIFF
--- a/manifests/object/host.pp
+++ b/manifests/object/host.pp
@@ -45,6 +45,7 @@ define icinga2::object::host (
   $target_file_mode = '0644',
   $refresh_icinga2_service = true,
   $zone = undef,
+  $command_endpoint = undef,
 ) {
 
   #Do some validation of the class' parameters:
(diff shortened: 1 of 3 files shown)


arioch:
Thanks for the contribution! :+1: 

Could you rebase (fixup) that into a single commit?
And if possible please also add a spec test for this parameter to [spec/defines/icinga2_object_host_spec.rb](https://github.com/Icinga/puppet-icinga2/blob/develop/spec/defines/icinga2_object_host_spec.rb).

hashjo:
OK. Commits fixed and spec test added. Thanks

arioch:
Thx!
Merged upstream.

lazyfrosch:
@arioch btw. you can leave the PR open, and let the merge/push auto-close the PR, since you are pushing the PRs commits

arioch:
@lazyfrosch I don't merge PR's on Github as it causes a spaghetti timeline. I prefer a flat timeline. 
Sadly when manually merging patches it doesn't always close the PR.

PR #160, opened 2015-11-21
correction graphite + add new params, which included in icinga 2.4

Reamer:
I made a mistake in graphite feature class last time. This commit should correct it.

With Icinga2.4 we have new params in graphite object - add support.

Write some spec tests, that this mistake will not happen again.

### DIFF
--- a/manifests/object/graphitewriter.pp
+++ b/manifests/object/graphitewriter.pp
@@ -10,16 +10,34 @@
 #
 
 define icinga2::object::graphitewriter (
-  $host       = '127.0.0.1',
-  $port       = 2003,
-  #Put the object files this defined type generates in features-available
-  #since the Graphite writer feature is one that has to be explicitly enabled.
-  $target_dir = '/etc/icinga2/features-available',
-  $file_name  = "${name}.conf",
+  $host                   = '127.0.0.1',
+  $port                   = 2003,
+  $host_name_template     = undef,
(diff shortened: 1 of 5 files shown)


arioch:
Could you split adding the new parameters to a new commit?
Other than that it looks good to me.
Also bonus points for adding the rspec. :+1:

Reamer:
I split the commit.

Greetings Reamer

Reamer:
Do you need any thing else?

manifests/object/graphitewriter.pp
```diff
@@ -10,17 +10,35 @@
...
+  $host_name_template     = undef,
+  $service_name_template  = undef,
+  # Only avaiable in icinga2 >= 2.4
+  $enable_send_thresholds = undef,
+  $enable_send_metadata   = undef,
+  # This will be only avaiable in some icinga 2 versions for examlple 2.4
+  $enable_legacy_mode     = undef,
+  # Put the object files this defined type generates in features-available
+  # since the Graphite writer feature is one that has to be explicitly enabled.
+  $target_dir             = '/etc/icinga2/features-available',
```
lazyfrosch, line 24: This should not be the default value here, but set when called from the feature.

lazyfrosch:
Merged, with the addition of the default path fixed.

Thanks @Reamer !

PR #161, opened 2015-11-23
README: db_password param was changed to db_pass

arnd:
The db_password parameter was renamed, but not all
references to it in the documentation (README.md).

This patch fixes that.


arioch:
Thanks for catching this one. 
Merged upstream. :+1:

PR #162, opened 2015-12-10
Add tagged string to permit using raw string instead of to_dump

nerzhul:


lib/puppet/icinga2/utils.rb
```diff
@@ -18,6 +18,8 @@ def self.config_value(value, indent=2)
               return value.to_i
             elsif value.match(/^\d+\.\d+$/)
               return value.to_f
+	    elsif value.match(/^%@@.*%@@$/)
```
Farzy, line 21: What about '@@' instead of '%@@'?
You should also write a test and update the documentation.

lazyfrosch:
Interesting idea... Though I'm not really sure if I like it.

nerzhul:
@larsengels how can we use a variable easily if we don't have a pattern ? You module is very generic and dump variables, which is good for security but not good for loops which are a very very useful case in icinga2. I don't think there is a problem if loop over hash, but loop over array has a problem

lazyfrosch:
@nerzhul there is also a way to deploy additional configs without using the defines.

I usually recommend to do something like a catalog for complex services or additional object definitions.

So that Puppet can be used for deploying hosts and application specific service definitions.

nerzhul:
@lazyfrosch you mean using exported resources for apply_service too ? it's not good for our usage. Puppetdb is very used.

lazyfrosch:
@nerzhul usually i do this via a `file` or `vcsrepo` resource on the master:

``` puppet
file { 'icinga catalog':
  ensure  => directory,
  path    => '/etc/icinga2/zones.d/master/catalog',
  source  => 'puppet:///icinga2_user/catalog',
  recurse => true,
  force   => true,
  purge   => true,
}
```

Or a `vcsrepo` that checks out that dir.

Makes sense? I plan to document this when rewriting the README

nerzhul:
@lazyfrosch our approach is to export all our hosts to the puppetdb catalog with tags. Icinga2 master nodes fetch hosts from puppet using exported resources (we configure our DNS using this method).
For sensors, they are deployed globally on all the icinga2 master nodes and we use NRPE to monitor (we have Debian from 4 to 8) their, using sometimes some arguments in the commands.
In the example covered by this PR, i have a puppet fact which collect all mounted partitions and icinga2 add them into icinga::vars::mounts . Then we for a apply_service for (mount_path in host.vars.mounts)
it's a proper way to collect and make disk sensors flexible because each server has different partition scheme depending on its role. The only problem is to use properly the variable without doing static thing. Here we are totally flexible and don't searching to define static sensors

lazyfrosch:
Yes, this is the usual behavior and pretty okay.

My so called 'catalog' only contains:
- apply "rules"
- commands
- notification settings

You cannot map the whole feature set of apply via Puppet settings. So this is the best way to do so.

The config files can be included in a Puppet module (files), or with a VCS repo you control somewhere.

nerzhul:
Now i see what are you talking about.
Our puppet architecture is hiera centric. Adding a puppet static file is avoided when we can. It's why i added this patch

Markus Frosch notifications@github.com a écrit :

lazyfrosch:
I get that, but do you really want to put anything Icinga 2 can do in its language, including Service for loops, if blocks, and what not, into the Puppet abstraction? That is insane.

nerzhul:
For my usage case if conditions aren't needed but loops with array variables yes. The implémentation is trivial unlike if conditions.
In our datacenter (400 machines) everything is fully managed with puppet (except business apps which are deployed and built by ansible)

Markus Frosch notifications@github.com a écrit :

lazyfrosch:
Then we should probably find a way to set custom config via something like hiera, instead of doing magic inside parameter values.

A proper vcs repo with checked-in config files is still a pretty safe and replicable way to define such config.

nerzhul:
@lazyfrosch why not but for this precise case i think magic value is a simple thing :)

lazyfrosch:
@arioch your opinion on this?

lazyfrosch:
I'm sorry to close this, but everywhere we use the `icinga2_config_value` function, we don't want the user to inject some magic code.

Some documentation for my "catalogue" should be done: https://dev.icinga.org/issues/10951

PR #164, opened 2015-12-16
Hotfix objects apply service

moritzfriedrich:
fixed call apply_service_to_host -> apply_service

Perhaps the hiera nameing needs to be adjust.


lazyfrosch:
Thanks for your input, but the PR looks broken, the final diff is rather short.

Is the final diff the only thing you wanted to change?

moritzfriedrich:
Oh yes, the finall diff is the only changed part. The other stuff was already pushed by someone else.

lazyfrosch:
Ah, good to know.

lazyfrosch:
Though we should also use `icinga2::object::apply_service` in hiera... I'll deprecate the old variable.

PR #167, opened 2016-01-11
WIP: Refactor hiera bindings

lazyfrosch:
Hiera hashes are not pulled in by auto-loading, and not usage of hiera_ functions.

fixes #9148

Please review.

Notes:
- icinga2::objects is gone
- icinga2::object is auto loaded


lazyfrosch:
@arioch poke :wink:

jkroepke:
Why do you remove the hiera_ functions?

lazyfrosch:
Because it is not best practice, which is using the auto binding feature im using now.

See https://dev.icinga.org/issues/9148 and https://dev.icinga.org/attachments/download/2419/hiera.png

jkroepke:
Normally, hiera merge does not work with autobindings.

Markus Frosch notifications@github.com schrieb am Di., 12. Jan. 2016,
10:10:

lazyfrosch:
I understand that, but if you need to use merging hashes, you should probably not rely in the base module to do it, but realize this via a custom additional module.

Reamer:
Mayflower/php has also approved status (https://forge.puppetlabs.com/mayflower/php/compatibility)
But has hiera_hash in init.pp 
https://github.com/mayflower/puppet-php/blob/master/manifests/init.pp

I am a fan of hiera_hash in manifests/objects.pp. Can we keep this class in this module, but default disabled?

lazyfrosch:
@Reamer thats interesting, wondering what @arioch 's opinion is :D

jkroepke:
elasticsearch/elasticsearch is approved and use hiera_merge for hash
params, too.

Viele Grüße,
Jan-Otto Kröpke

2016-01-12 12:18 GMT+01:00 Markus Frosch notifications@github.com:

lazyfrosch:
Okay I get your guy's points, fair enough. Thanks for the input :smiley: 

Let's just cleanup that class and leave it as it is. Closing this PR

arioch:
Straight from the horse's mouth...

hiera

TL;DR: A **Puppetlabs employee's** module was declined solely because of hiera_*() calls in his module.
Hence I still think it's a bad idea.

lazyfrosch:
@arioch are you sure? It seems more like how the module "types" is actually wrapping around the resource types of Puppet.

The Puppetlabs guys answer is irritating...

lazyfrosch:
I'd say we'll see later.

PR #168, opened 2016-01-15
Change a default parameter of the host object.

Ziaunys:
- This changes the target_file_name parameter of the host object to use
  the $name variable instead of the $fqdn fact to be consistent with the
  other object define types.
- The test hasn't been updated since the results will have the same
  results.


lazyfrosch:
Thanks for catching this Eric!

PR #169, opened 2016-01-17
Added custom parameter for apply_service

jkroepke:
Hi,

I want use services like this:

```
apply Service "check_http" for (name => config in host.vars.http) {
      import "generic-service"
  display_name = "HTTP " + name + ""
  assign where "linux" in host.groups
  check_command = "check_http"
  vars += config
}
```

Currently, the line "vars += config" is impossible to set via puppet.

### DIFF
--- a/spec/defines/icinga2_object_apply_service_spec.rb
+++ b/spec/defines/icinga2_object_apply_service_spec.rb
@@ -1,0 +1,85 @@
+require 'spec_helper'
+require 'variants'
+
+describe 'icinga2::object::apply_service', :type => :define do
+  IcingaPuppet.variants.each do |name, facts|
+    context "on #{name} with basic parameters" do
+      let :facts do
+        facts
+      end
+      let :pre_condition do
+        " include 'icinga2'"
+      end
+
+      let(:title) { 'check_http' }
+
+      object_file = '/etc/icinga2/objects/applys/check_http.conf'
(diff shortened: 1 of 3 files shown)


lazyfrosch:
I like the idea, will try it out and merge.

Could you provide a simple spec test?

spec/defines/icinga2_object_apply_service_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+      it { should_not contain_file(object_file).with_content(/^\s*check_command = "check_http"/) }
+
+    end
+    context "on #{name} with custom_prepend parameter in string format" do
+      let :facts do
+        facts
+      end
+      let :params do
+      {
+        custom_prepend: => 'vars += config',
```
lazyfrosch: Shouldn't commend without reading closely :tired_face:

`:name => 'value',`

lazyfrosch:
You can run the tests locally, e.g.

``` bash
rake spec_prep
rspec spec/defines/icinga2_object_apply_service_spec.rb
# or
rake spec
```

jkroepke:
Local test do not helped me. This is the result, if I run local tests:

679 examples, 149 failures, 11 pending

Anyway, I wrote working tests.

lazyfrosch:
@jkroepke thanks for the awesome test! Only one remark, usually you never test for breakage, just for the sake of the validate_\* command to work :)

PR #171, opened 2016-01-22
added hostgroup documentation

mzhaase:
hostgroup documentation based on host documentation.


lazyfrosch:
Thanks, I've fixed format and a missing single quote

PR #173, opened 2016-02-11
Add support for OpenBSD

buzzdeee:
- Add OpenBSD case statement to params.pp
- Change all *target_file_groups from 'root' -> '0', at least on *BSD, there is no group 'root', but is called 'wheel'. Using group '0' instead of a hardcoded name, is a common pattern.  That should help in the future with other *BSD as well. Alternatively there could be something like a OS dependent $root_group in params.pp, and referencing that all over the place. Let me know if you would prefer it that way.

With the following module variables in Hiera, I get a very basic Icinga up and running:

```
icinga2::manage_repos: false
icinga2::install_mailutils: false
icinga2::pid_file: '/var/run/icinga2/icinga2.pid'
```

Alternatively, those variables could be made OS dependent, but for the time being, didn't wanted to go that far (:

### DIFF
--- a/manifests/checkplugin.pp
+++ b/manifests/checkplugin.pp
@@ -7,7 +7,7 @@ define icinga2::checkplugin (
   $checkplugin_name                     = $name,
   $checkplugin_libdir                   = $::icinga2::checkplugin_libdir,
   $checkplugin_target_file_owner        = 'root',
-  $checkplugin_target_file_group        = 'root',
+  $checkplugin_target_file_group        = '0',
   $checkplugin_target_file_mode         = '0755',
   $checkplugin_file_distribution_method = 'content',
   $checkplugin_template_module          = 'icinga2',
--- a/manifests/conf.pp
+++ b/manifests/conf.pp
@@ -26,7 +26,7 @@ define icinga2::conf (
   $target_dir        = '/etc/icinga2/conf.d',
   $target_file_name  = "${name}.conf",
   $target_file_owner = 'root',
-  $target_file_group = 'root',
+  $target_file_group = '0',
   $target_file_mode  = '0644',
 ) {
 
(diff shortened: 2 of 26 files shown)


ChrisHeerschap:
+1 for OpenBSD - looks like the same code would work for FreeBSD too, no?

buzzdeee:
I don't have a FreeBSD around, but at least the group change from 'root' -> '0' should help. I guess with regard to the params.pp changes, FreeBSD would need a separate entry. I would guess the plugin_packages have different names, as well as $config_group and maybe the $checkplugin_libdir.

ChrisHeerschap:
Yep, I had just checked monitoring-plugins - that's the same, but now that I look for the others, they're not there.

buzzdeee:
On OpenBSD monitoring-plugins is the main package, the others are subpackages, you usually only install, when you need them, because they pull in otherwise not needed dependencies.
I'm open to suggestions (:, and could reduce the list down to just monitoring-plugins as default $plugin_packages. If someone really needs, he can override the parameter, or if I see that right, use $plugin_packages_extra parameter when including icinga2.

zachfi:
Woo!

zachfi:
For FreeBSD, or others, the `plugin_packages` variable could just be set undef, and wrap a conditional around https://github.com/Icinga/puppet-icinga2/blob/develop/manifests/install.pp#L13 before attempting to install, just to make sure the variable is set.  Seems simple enough.  I'd love to se both FreeBSD and OpenBSD support added, so big +1 here.

buzzdeee:
I changed params.pp for OpenBSD the $plugin_packages array to only contain the one basic monitoring-plugins package instead of the full-blown list. Whoever needs more, is free to add, but I think it makes more sense to have the default list rather small.

zachfi:
+1  I would merge if I could.

ChrisHeerschap:
+1 (again) - what @xaque208 said. Merge worthy in my mind.

arioch:
Excellent, thank you @buzzdeee!
Also special thanks to @ChrisHeerschap and @xaque208 for reviewing.

Merged upstream. :+1:

PR #174, opened 2016-02-18
Changes to the content of a feature.conf file should trigger a

buzzdeee:
restart of icinga

While changing the bind_host parameter to listen on IPv6 instead of IPv4 of the already enabled api feature, I recognized that the file /etc/icinga2/features-available/api.conf successfully got updated, but the service was not restarted.

This little change now triggers a restart of Icinga on feature parameter changes.

I test for if $manage_file == true because the same check is used a bit earlier in that file as well. Alternatively, I could have added the trigger a couple of lines up within the already existing 'f $manage_file == true', and there guarding it with the if $::icinga2::manage_service ... but found this place a bit more appropriate.

let me know if there is something to change/enhance/etc... otherwise please consider for merging 

cheers,
Sebastian

### DIFF
--- a/manifests/feature.pp
+++ b/manifests/feature.pp
@@ -18,7 +18,7 @@ define icinga2::feature (
     mode  => $::icinga2::config_mode,
   }
 
-  if $manage_file == true {
+  if $manage_file {
 
     if $content {
       $content_rel = $content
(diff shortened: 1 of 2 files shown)


arioch:
It's better to use `if $manage_file {` instead of `if $manage_file == true {`.
Could you also provide a spec test for this?

Other than that it's good to merge. :+1:

buzzdeee:
then I'll also change the test above to be only if $manage_file {
with regard to the spec test, will see that I find some time on the weekend hopefully

buzzdeee:
changed both occurences of that 'if' statement. Adding a spec test took a while to get right, but eventually figured it out. Added tests for both cases of $manage_file.

cheers

PR #177, opened 2016-03-09
Add option to notify service, if needed.

aschaber1:
This is useful, if $::icinga2::package_ensure is set to latest.


arioch:
Looking good, merged upstream. :+1:

PR #179, opened 2016-03-09
WIP: FreeBSD support

zachfi:
This work takes a first pass over adding support for the FreeBSD
platform.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -115,6 +115,20 @@ class icinga2::params {
       # Icinga2 client settings
       $checkplugin_libdir   = '/usr/local/libexec/nagios'
     }
+    'FreeBSD': {
+      $plugin_packages = [
+        'monitoring-plugins',
+      ]
+      $mail_package = undef
+
+      #Settings for /etc/icinga2/:
+      $config_owner = 'icinga'
+      $config_group = 'icinga'
+      $config_mode  = '0640'
+
+      # Icinga2 client settings
+      $checkplugin_libdir   = '/usr/local/libexec/nagios'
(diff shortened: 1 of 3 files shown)


zachfi:
The configuration directory here for icinga2 on FreeBSD should be `/usr/local/etc/icinga2`, but currently its not configurable.  How would folks like this to be addressed?  I'm happy to make the config directory tunable, but its hardcoded in lots of places currently, so before I embark on such a task, I'd like to see if there is a preferred way to handle this.  I know the defaults everywhere else are for `/etc/icinga2`, but since we don't deploy all files we include, the service remains broken on FreeBSD.

zachfi:
@arioch @lazyfrosch Any thoughts here as to adding support for FreeBSD?

lazyfrosch:
@xaque208 if you have a proper test setup, go ahead!

:+1: for config_dir support

arioch:
What's the current status of this WIP?
I see some overlap with #196. Does this cover what you need?

zachfi:
Once the prefix stuff is in place, I can begin adding support.  On my personal infrastructure, I've symlinked the directories and the module works quite well, so all that is required for FreeBSD support is:
- remove any root group use and replace with '0'
- Any configuration paths prefixed with `/usr/local`
- Tests to keep this from regressing

I think thats it really.  I'll take a pass over #196 again.

Puppet4 native data in modules would make this quite a bit simpler.  Did you see my report I filed about it on the ticket tracker?

zachfi:
Looks like the module is being rewritten.  No need to keep this open.

PR #180, opened 2016-03-12
Missing Quotation

dagobert:
The missing quotation of cmd results in not being able to use ones own path. It has to be changed to the same as in checkcommand template.


arioch:
I've fixed the broken tests.
If you'd like to learn more about those check b1fa132bf4ac37a755d2a3e19c5ca8ed250aee1a.

Merged upstream.

PR #181, opened 2016-03-13
Added OracleLinux as a supported OS

TheFlyingCorpse:
Added support for Oracle Linux.


arioch:
Looking good!
Merged upstream.

PR #184, opened 2016-03-22
better speaking display_names for services

janvonde:
This is a simple fix for https://dev.icinga.org/issues/11358

### DIFF
--- a/manifests/object/apply_service.pp
+++ b/manifests/object/apply_service.pp
@@ -14,7 +14,7 @@ define icinga2::object::apply_service (
   $object_servicename = $name,
   $apply = 'to Host',
   $templates = ['generic-service'],
-  $display_name = $name,
+  $display_name = 'name',
   $assign_where = undef,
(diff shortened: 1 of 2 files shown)


arioch:
What's the benefit of adding just 'name' to every service's display name?

dgoetz:
I think the default should be name (only name will be resolved to the name internally by icinga2) or simply not setting the display_name at all but the more important part of the commit is removing the quotes from the template allows to do some magic with the display_name in apply for like "Interface " + interface or "Disk " + disk + " Usage".

dnsmichi:
Imho the default should not be set although it is considered best practice to define one. When using apply rules you can access the current context as well as referenced objects (e.g. a service apply allows you to use host attributes). That way you can modify certain attributes being set inside the apply rule. That is helpful for the display_name but also allows you to use conditions (if (match("foo*"), host.name) { } ...). In terms of the puppet module, one could easily use string concat to generate their own display_name.

PR #186, opened 2016-03-23
Ensure apt update occurs before package installation

erikanderson:


erikanderson:
Will rework this and do another pr

bwolf:
I don't like this. IMHO this should be solved somewhere else. But maybe I'm wrong.

I Use the following in my base/profile to ensure that apt has the recent package db:

``` puppet
Exec['apt_update'] -> Package <| |>
```

arioch:
I agree with @bwolf, this should not be handled by the module.

PR #187, opened 2016-03-25
Remove notify_service in favor of manage_service.

aschaber1:
Using the already established manage_service seems like a better idea.


arioch:
Excellent. 👍 
Merged upstream.

PR #188, opened 2016-03-28
Fix inheritance of config owner/group/mode for the configuration files

TheFlyingCorpse:
Made it easier to inherit config owner/group/mode to configuration files from the globally set parameters.


zachfi:
+1 though I don't see the modification to init.pp.

TheFlyingCorpse:
@xaque208 - I did the modification that was already started in a few files under "manifests/object/", now its consistent in that it will inherit what is used. I dont change init because it sets the parameters based on what flavour the OS is. The intent with this change is to make it similar for all the created files that it inherits from the root class unless otherwise specified.

TheFlyingCorpse:
Still waiting for this to be merged.

arioch:
Excellent. 👍 
Merged upstream.

PR #189, opened 2016-03-28
Feature/windows supportv5

TheFlyingCorpse:
Adds support for Windows to Puppet.

Note that the 2 facts added are:
- i2_dirprefix: prefix for where Icinga2 is installed, this is prefixed to any path, if its empty or undefined it doesnt matter, then its the paths as usual (like linux). Could be useful if one decides to prefix the icinga2 install on linux later on.
- i2_puppet_facts: adds a ssldir fact, without this there is no support for Puppet3 that I can see. This is because the ssldir is based on the master and not what the agent has. With this fact, it works regardless of Puppet3 or 4, it also makes sure it works with Windows.


TheFlyingCorpse:
Another note: I havent been able to figure out how to make Windows install puppet via chocolatey (or other sources) as part of the include class, thus I've done this hack instead:
`if $::kernel == 'windows' {
    include chocolatey
    Package { provider => chocolatey, }
    if $::i2_dirprefix {
      class { 'labdomain::icingaclient': }
    } else {
      package { 'icinga2':
        ensure => latest,
      }
    }
  } else {
    class { 'labdomain::icingaclient': }
  }`

For the SSL, I use this (didnt dare include it as default, awaiting feedback from this PR:
`$ssldir = $facts['i2_puppet_facts']['ssldir']
  class { '::icinga2::pki::puppet':
    ca_source   => "$ssldir/certs/ca.pem",
    cert_source => "$ssldir/certs/${::fqdn}.pem",
    key_source  => "$ssldir/private_keys/${::fqdn}.pem",
    crl_source  => "$ssldir/crl.pem",
  }
`

Windows itself is configured with this for the base icinga2 class:
`class { '::icinga2':
      install_plugins => false,
      install_mailutils => false,
    }
`

TheFlyingCorpse:
I did an --amend to repush it, it failed setting up the verification environment.

TheFlyingCorpse:
Resolved issue of package installation on windows, its handled via the "manage_repos", if set to false it assumes other source of package.

The $ssldir is still something I'd like feedback on if it should be internal to module or not.

arioch:
Thanks for this huge effort. It's very much appreciated. :+1:

Is it safe to assume this is still a work in progress?

Also would you be available for testing patches against this platform? I can't speak for everyone in the team, but I personally don't have the resources to test any windows related stuff.

lib/facter/i2_dirprefix.rb
```diff
@@ -0,0 +1,34 @@
+Facter.add(:i2_dirprefix) do
```
zachfi, line 1: Why are there two facts by the same name in this file?  Is that intentional?

zachfi:
This work looks pretty good to me.  It will also help any other platforms easily implement a prefix that need it.  Thanks for the effort.

TheFlyingCorpse:
@arioch - the code as it is now works ok. I'm working on improving the icinga2 installer for windows to make it better suited for the future (64 bit installer, likely msi instead of exe based. + looking for ways to make it to a format required for Server 2016 Nano called .appx).
If/When I get the MSI done, I can redo some small parts in here to make it cleaner. 

What I'd really like some help with is how to make the spec work with Windows. I've tried for a few days locally and against the Travis-CI on this  PR, but I cant seem to figure it out. Thoughts? 

I'm on IRC most evenings EU time if anyone of you want to give me feedback on this directly or drop me some hints :+1:

lib/facter/i2_dirprefix.rb
TheFlyingCorpse, line 1: For now it is because the installer for Windows is 32bit, on 64bit systems the registry key is different than where it is on 32bit because of this, the key is vital to figuring out where icinga2 is installed.

zachfi:
What can we help with here?  I'd been holding out on adding FreeBSD support that also uses a different config dir prefix until this one had been merged, but I see now its sitting a bit.

TheFlyingCorpse:
@xaque208 - This has been sitting as I was waiting for modifications to the icinga2 package on Windows. It seems to be nearly there so I am going to modify the commit to match it.
I plan on having 4 prefixes:
- sbinprefix
- varprefix
- etcprefix
- shareprefix

What I'd need some help with is how to go forward with the spec tests for Windows, I can't say I understand how to make them work properly.

zachfi:
@TheFlyingCorpse I just looked over a couple of failures in the test.  To start, I think amending the dependencies in metadata.json, then also adding chocalately to the `.fixtures.yml` file should at least allow the class to be found for testing.  Then we can work out whats next.

zachfi:
@TheFlyingCorpse I'd also be in support of breaking up the commits a bit here to help them get merged.  The first commit in this passed, and is good work by itself, so to get it merged, it might be worth submitting on its own.  Just thinking out loud, and I'm not a committer here :)

TheFlyingCorpse:
I'll see how its logical to split this up, perhaps as following:
1) The prefix logic + amendments to prefixes in all relevant files & specs
-- Prefix logic:
--- Generic prefix for everything, 
--- Alt: specific prefix per type as mentioned before.
2) Windows support, this is a "trivial" change now that the prefix is taken out.

I have a PR open already to fix the inheritance of config file owner/mode.
Suggestions welcome if you would see this split up more.

TheFlyingCorpse:
Two commits done today that are separate PR's:
- #196 is the prefix paths
- #197 is to use the local puppet agents environment in place of the puppetservers, now it will work with puppet 3.x agents against a 4.x master.

I will rebase and work on the windows version to be based on these two shortly.

TheFlyingCorpse:
Closing this as it will be redone with a fresh codebase.

PR #190, opened 2016-04-09
Documentation example quote fix

zachfi:
The template no longer appears to need the extra quotes to reproduce the
default mail command.  This work simply removes the extra quotes that
caused a syntax error, now resulting in a working configuration.


zachfi:
Note the quotes in the template here: https://github.com/Icinga/puppet-icinga2/blob/develop/templates/object/notificationcommand.conf.erb#L13

arioch:
Thanks for catching this.
Merged upstream.

PR #191, opened 2016-04-11
Fix apply_service.conf.erb indentation

n00by:
Trim some unnecessary preceding indentation.


arioch:
Excellent, thanks.
Merged upstream.

PR #193, opened 2016-04-18
Fix checkcommand manifest and template

n00by:
Update the validation of the parameters in the checkcommand manifest. Remove the obsolete
"if" statement which prevents necessary parameter validations. The parameter "command"
can either be an array or a string.

Fix the template. Make sure the "command" parameter is passed as an array to the
"each_with_index" method, otherwise an exeption is thrown. If the "command" parameter
consists of an array of individual command arguments, don't add the command path to
each element of the array. This should only be done for the first element of the
array (the command).

This fixes establish the desired behavior for the "command" parameter as described in
the icinga2 docs.

Update the rspec tests.

### DIFF
--- a/manifests/object/checkcommand.pp
+++ b/manifests/object/checkcommand.pp
@@ -13,7 +13,7 @@ define icinga2::object::checkcommand (
   $command,
   $object_checkcommandname = $name,
   $templates                             = ['plugin-check-command'],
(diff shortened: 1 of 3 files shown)


arioch:
Excellent! 👍 
Merged upstream.

PR #194, opened 2016-04-20
update zone documentation. enpoints are no arrays, they are hashes now

rwaffen:


gunnarbeutner:
The 'endpoints' attribute is actually _not_ a dictionary. It's an array and apparently the Icinga 2 documentation incorrectly states that it's a dictionary.

rwaffen:
Than why is there a validate_hash and create_resource in the code. I didn't get it to work with arrays.

manifests/object/zone.pp

```
  if $endpoints {
     validate_hash($endpoints)
     $_endpoints = keys($endpoints)

     create_resources('icinga2::object::endpoint', $endpoints)
}
```

dgoetz:
Sorry, guys, you are talking apples and oranges.

In Icinga 2 the attribute endpoints of the zone object is an array, but in the puppet module it is a dictionary to provide additional attributes to also create the endpoint objects automatically.

rwaffen:
oh okay, that's the point... was wondering how that would be an array in puppet when it is checked that it has to be a hash 😄

arioch:
Excellent. 👍 
Merged upstream.

PR #195, opened 2016-04-22
Fix icinga_ca_port usage in `icinga2 pki` command

ngoduykhanh:
It looks like we used wrong variable in the `icinga2 pki` command. Bellow is my error:

```
Error: icinga2 pki request --host 'masterserver.local'5665 --ca '/etc/icinga2/pki/ca.crt' --key '/etc/icinga2/pki/satellite1.local.key' --cert '/etc/icinga2/pki/satellite1.local.crt' --trustedcert '/etc/icinga2/pki/trusted-cert.crt' --ticket '243de7fd4578bda368af92a4d81d058a6d92a592' returned 1 instead of one of [0]
```

This PR is for fixing that error. Please help to review it :)

### DIFF
--- a/manifests/pki/icinga.pp
+++ b/manifests/pki/icinga.pp
@@ -50,14 +50,14 @@ class icinga2::pki::icinga (
   } ->
 
   exec { 'icinga2 pki get trusted-cert':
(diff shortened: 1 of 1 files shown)


arioch:
That's a bug that went unnoticed due to lack of test coverage.
I've merged your PR and adjusted the spec in  77e13cd0598673347697c456f1ff895c35a41aec

Thanks for catching and fixing this! 👍

PR #197, opened 2016-04-23
ssldir from puppet agent

TheFlyingCorpse:
This fixes the dependence on having the same puppet version (3 vs 4) on all agents AND puppetserver. 
With this you can have legacy agents (3.7 that I have tested) with puppetserver 4.x.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -132,6 +132,14 @@ class icinga2::params {
     $var_dir = '/var'
   }
 
+  # Use the fact for the local agents ssldir, if it is not found it will use the masters.
+  if $::icinga2_puppet_ssldir {
+    $puppet_ssldir = $::icinga2_puppet_ssldir
+  }
+  else {
+    $puppet_ssldir = $::settings::ssldir
+  }
+
   # the schema is now using a set parameter for the os or the default.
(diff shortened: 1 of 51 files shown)


TheFlyingCorpse:
This commit implies that PR #196 is accepted.

manifests/params.pp
```diff
@@ -27,15 +27,53 @@
...
+  }
+  else {
+    $varprefix = '/var'
+  }
+
+  if $::i2_puppet_ssldir {
+    $puppetssldir = $::i2_puppet_ssldir
+  }
+  else {
+    $puppetssldir = $::settings::ssldir
```
zachfi: This is a poor default here because it will retrieve the master's ssldir, and not the agents.  This would only work in an environment where the master and the agent are the same operating system.  This should be at least overridable by the user as a parameter on a class, which is not possible given the current implementation.

TheFlyingCorpse: The fix in this is actually what you describe, it will use the agents
ssldir. Currently it does not use it, it uses the masters ssldir!

TheFlyingCorpse:
Will update this as often as I can based on the updates to PR 196, for now I need at least one more fix to 196 before this will be updated again.

Reamer:
:+1: also tested in my environment. Thank you.

TheFlyingCorpse:
Rebased

PR #198, opened 2016-04-25
Use correct alignment

zachfi:
Without this change, the alignment of the rendered template is incorrect when the user specifies a custom_append.


arioch:
Excellent, thanks!

Merged upstream.

PR #200, opened 2016-04-28
Fix and extend docu

rwaffen:
- fix docu, there was an open tag
- add hint on how to use the puppet external ca


arioch:
Cool, thanks! 
Merged upstream.

PR #201, opened 2016-05-01
Fix the checkcommand template vars attribute handling

n00by:
so the vars are written correctly into the CheckCommand object configuration. Before
the fix the template produced the following CheckCommand object configruation for
the vars attribute:

  vars += {
    vars.http_address = "$address$",
    vars.http_ssl     = false,
    vars.http_sni     = false
  }

instead of the desired

  vars.http_address = "$address$",
  vars.http_ssl     = false,
  vars.http_sni     = false


arioch:
Excellent!
Merged upstream.

PR #204, opened 2016-05-02
fix documentation for install_mailutils

LorenzBischof:
The documentation for the parameter to install `mail` packages was wrong. It also wasn't clear if the default is true or not.

### DIFF
--- a/manifests/params.pp
+++ b/manifests/params.pp
@@ -21,9 +21,6 @@ class icinga2::params {
   #Whether to install the plugin packages when the icinga2 class is applied:
   $install_nagios_plugins = true
 
-  #whether to install packages that provide the 'mail' binary
-  $install_mail_utils_package = false
-
   $config_template = 'icinga2/icinga2.conf.erb'
   $manage_service = true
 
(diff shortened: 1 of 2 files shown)


manifests/params.pp
```diff
@@ -21,9 +21,6 @@
   #Whether to install the plugin packages when the icinga2 class is applied:
   $install_nagios_plugins = true
 
-  #whether to install packages that provide the 'mail' binary
-  $install_mail_utils_package = false
```
zachfi, line 25: This isn't a documentation change as the commit suggests.  What is this change here to address?

LorenzBischof, line 25: It seems to be leftover code from a rename of the parameter. It doesn't get used anywhere else.

zachfi, line 25: This change should be a separate commit in my opinion.

LorenzBischof, line 25: I agree. I'll seperate it, when I'm at a computer.

LorenzBischof:
Ok, I seperated my changes into two commits.

zachfi:
+1 Would merge.

PR #205, opened 2016-05-03
Clean up database documentation and add tests

prozach:
Without this work, the documentation around setting up an initial server
is unclear.  This patch addresses this by fleshing out the documentation
after reviewing the code carefully to discover what the actual behavior
is.  Tests are also added here to ensure that what the documentation
says is true, is actually true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,45 +132,45 @@ icinga2::conf { 'baseservices':
 
 ###Server usage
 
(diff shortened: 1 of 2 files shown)


prozach:
Hi.

arioch:
Excellent. 👍 
Merged upstream.

prozach:
Yay, thank you @arioch.

PR #206, opened 2016-05-03
Fixes quotes in eventcommand template.

aschaber1:


zachfi:
How about a test for this?

aschaber1:
I'm not even sure, if this should stay this way, when I look at the checkcommand template.

PR #208, opened 2016-05-10
Add custom_prepend and custom_append parameter to object/host.pp and

buzzdeee:
the accompanying template. Idea taken from the apply_service
class/template.

This allows me to get around the limitation I faced with regard to apply_service
and the disks, see my comment here:
https://dev.icinga.org/issues/10804


arioch:
Looking good so far...
Could you also add a spec test to cover this? And maybe a little documentation to README.md as well?

buzzdeee:
good to know it looks good, ;) will look into spec tests/documentation as soon as possible

buzzdeee:
alright, how about this then

buzzdeee:
just rebased onto latest develop

buzzdeee:
ping
rebased for your convenience...

arioch:
Looking good. 👍 
Merged upstream.

PR #213, opened 2016-07-07
Fix notificationcommand template

n00by:
Make sure the "command" parameter is passed as an array to the
"each_with_index" method, otherwise an exeption is thrown.

Minor syntax corrections.


arioch:
Could you also adjust the spec tests?

n00by:
done

n00by:
@arioch please bump

arioch:
Thx @n00by, very much appreciated as always. 👍

PR #214, opened 2016-07-07
fix puppet lint config

n00by:
no-80chars option is no longer supported
use no-140chars-check option instead


arioch:
Only if you update puppet-lint to version 2. ;)
Updated while merging this PR.

Thanks!

PR #215, opened 2016-07-11
Add support for Amazon Linux

rwaffen:
- Add support for Amazon Linux
- Use EPEL 6 Packages because Amazon Linux is based on RHEL 6
- Tested on a legacy system of my company with Amazon Linux 2015.03


arioch:
Looking good, thanks!
Merged upstream.

PR #217, opened 2016-07-23
Add InfluxdbWriter

TheFlyingCorpse:
Add InfluxdbWriter feature and object


TheFlyingCorpse:
If merged after #196, this needs two small edits, both references of target_dir.
I'll rebase and fix if #196 is merged before.

PR #221, opened 2016-08-23
Update apply_notification objects.

n00by:
This includes the apply_notification_to_service and
apply_notification_to_host object.

Make it possible to set a string for "users" and "user_groups". This is
necessary if you want to use variables as values, e.g.

users = host.vars.notification.mail.users
user_groups = host.vars.notification.mail.groups

Update the rspec tests.


n00by:
@arioch the spec tests with ruby 1.9.3 fail because "Gem::InstallError: json_pure requires Ruby version ~> 2.0."

This should be fixed with Reames commit: https://github.com/Icinga/puppet-icinga2/pull/219/commits

And not sure yet why Puppet version > 4 spec tests fail.

PR #222, opened 2016-09-13
fix(mysql): add logic to fix mysql schema load

jrodriguezjr:


Reamer:
The "creates" parameter should do this check. The additional check is not required.

